PR #22, opened 2017-03-21
subsuperscript

artragis:


packages/sub_super/inline.js
```diff
@@ -0,0 +1,64 @@
+function locator (value, fromIndex) {
+  const index = value.indexOf('^', fromIndex) == -1? value.indexOf('~', fromIndex):value.indexOf('^', fromIndex)
```
artragis: 100 car? zet sérieux?

packages/sub_super/inline.js
```diff
@@ -0,0 +1,64 @@
...
+  return index
+}
+
+function inlinePlugin (opts = {}) {
+  function inlineTokenizer (eat, value, silent) {
+    if(value.startsWith("~") && !value.startsWith("~ ")) {
```
artragis: single quotes nazi...

Issue #61, opened 2017-06-06
Blockquotes has no source


artragis:
It seems that remark does not handle `Source` assignment for blockquote element.

This code : 

```markdown
> blablabla
Source: blabla
```

Should be interpreted as "quote blablabla with source blabla" : 
![image](https://cloud.githubusercontent.com/assets/1523653/26815712/1d1cf6f4-4a8e-11e7-982f-23a84ca3cef5.png)

For now mdast tree is : 
```
─ blockquote[1] (10:1-11:16, 248-275)
│  └─ paragraph[1] (10:3-11:16, 250-275)
│     └─ text: "blablabla\nSource: Element" (10:3-11:16, 250-275)
```

PR #62, opened 2017-06-06
rebber blockquote

artragis:
This PR adds blockquote support for mdast->latex compilation.

This uses the `Quotation` environment as defined in [this line](https://github.com/zestedesavoir/latex-template/blob/76021abc62562a02628a40a3a52f07fbd80a07a1/zmdocument.cls#L160).

I did not use a "standard" quotation environment as latex has many standard to do that.

Be careful, this will trully work only if #61 is handled. For now I use "anonymous" placeholder for empty source.

### DIFF
--- a/packages/rebber/__tests__/fixtures/blockquote.fixture.md
+++ b/packages/rebber/__tests__/fixtures/blockquote.fixture.md
@@ -1,0 +1,5 @@
+> a quote
+
+> a 
+> multiline
+> quote
--- a/packages/rebber/__tests__/fixtures/blockquote.fixture.md
+++ b/packages/rebber/__tests__/fixtures/blockquote.fixture.md
@@ -1,5 +1,5 @@
 > a quote
 
-> a 
+> a
 > multiline
 > quote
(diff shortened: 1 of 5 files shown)


packages/rebber/src/types/blockquote.js
```diff
@@ -0,0 +1,12 @@
+/* Expose. */
+all = require('../all')
```
vhf: `const all = …`

artragis:
done @vhf

packages/rebber/__tests__/fixtures/blockquote.fixture.md
```diff
@@ -0,0 +1,5 @@
+> a quote
+
+> a 
+> multiline
+> quote
```
vhf: This fixture will get parsed as a single quote, not two, please fix the expected latex to reflect this.

Example: (test it on zds or https://zestedesavoir.github.io/zmarkdown/public/html.html)

```markdown
> foo

> bar
```

=>

> foo
> bar

Issue #44, opened 2017-04-16
implement grid_tables


artragis:
Gridtables are tables that we draw thanks to dash, equals and pipe. They fully support rowspan and colspan, can handle any block and markdown formatting.

Example : 

```markdown
+----------------------------+-----------------------+-----------------------+
| Col1                       | Col2                  | Col3                  |
+============================+=======================+=======================+
|Normal column               | col span column                               |
+----------------------------+-----------------------+-----------------------+
| Row span column            |  x                    |```text                |
|                            |                       | a code                |
|                            |                       |```                    |
+                            +-----------------------+-----------------------+
|                            | x                     | x                     |
+----------------------------+-----------------------+-----------------------+
```

![image](https://cloud.githubusercontent.com/assets/1523653/26816618/d9a08332-4a92-11e7-8196-fe1c2a199c86.png)

# implémentation de base

https://github.com/zestedesavoir/Python-ZMarkdown/blob/master-zds/zmarkdown/extensions/grid_tables.py

# les tests

- https://github.com/zestedesavoir/zmarkdown/blob/master/test/fixtures/zds/extensions/grid_tables.txt
- https://github.com/zestedesavoir/zmarkdown/blob/master/test/fixtures/zds/extensions/grid_tables.html

PR #65, opened 2017-06-08
implement grid_tables

AmarOk1412:
cf #44 

# Note

The test for
```
+---+-----------+
| A |           |
+---+-----------+
| B | C         |
+-------+---+---+
| D   E | F | G |
+-------+---+---+
```
changed because an useless column was created.
The generated HTML with `remark-grid-tables` gives the same result as before.

And the test for
```
+--- A ---+
|         |
+---------+
|         |
+---------+
```

Because
```
+---------+
|         |
+---------+
```
is a table of one cell.


packages/remark-grid-tables/src/index.js
```diff
@@ -0,0 +1,469 @@
...
+    this._lines = newLines
+  }
+
+  function extractTableContent (lines, linesInfos, hasHeader) {
+    const table = new Table(linesInfos)
+
+    for (let l = 0; l < lines.length; ++l) {
+      const line = lines[l]
+      const matchHeader = hasHeader & isHeaderLine(line) !== null
+      const isEndLine = matchHeader | isPartLine(line) !== null
```
vhf: Please add a comment explaining the use of these two bitwise operators (or refactor them to make it clear that you're actually working with booleans)

packages/remark-grid-tables/src/index.js
```diff
@@ -0,0 +1,469 @@
...
+
+      if (isEndLine) {
+        // It is a header, a plain line or a line with plain line part.
+        // First, update the last row
+        table.lastPart().updateWithMainLine(line, isEndLine)
+
+        // Create the new raw
```
vhf: row

packages/remark-grid-tables/src/index.js
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  function generateTable (tableContent, now, tokenizer) {
+    let wrapperClassName = 'table-wrapper'
+    if (classNames.wrapper) {
+      wrapperClassName = classNames.wrapper
+    }
+    // Generate the gridTable node to insert in the AST
+    const tableWrapper = {
+      type: 'element',
```
vhf: `element` is a [HAST](https://github.com/syntax-tree/hast) node, not a [MDAST](https://github.com/syntax-tree/mdast).

If we create `element` nodes, then this is a `rehype-` plugin -- it is working on HAST tree and not on MDAST tree.

`.data.hName` are HAST properties that can be defined on any HAST or MDAST nodes though, no need to change that.

I'm really in favour of having this plugin as `remark-` because we'll need these nodes in MDAST when we compile to LaTeX. I suggest the following: Since `Table` `TableCell` and `TableRow` exist in MDAST, use these node types instead of `element`. Get rid of the `tableWrapper`: it's rehype's job to wrap a table in a div if needed, the MD AST should know about this.

Keep all `node.data.h*` you currently have, I don't they'll get overridden by rehype. This way we would achieve:

* using MDAST node types
* having our nice `thead` and `rowspan` `colspan` etc

AmarOk1412:
@vhf Note: I didn't change 
```javascript
const tableElt = {
      type: 'gridTable',
      children: [],
      data: {
        hName: 'table',
      }
    }
```

to `type: 'table'` because of https://github.com/syntax-tree/mdast-util-to-hast/blob/master/lib/handlers/table.js

PR #68, opened 2017-06-09
[WIP]Support for sourced quotation

artragis:
This enables sourced quote. We need it for zds plugin.

This is wip as I need :

- a rehype plugin to add the data to html
- a bunch of test that truely work

note that for now, the second is my biggest problem.


vhf:
> a rehype plugin to add the data to html
I'd do everything in this plugin, no need for a rehype plugin.

Whenever you find a _source_, add what you need to it's `data` property. I'm pretty sure this should work.

vhf:
Here are fixtures for your tests:

* https://github.com/zestedesavoir/Python-ZMarkdown/blob/70cc535226d65bada95f4219aa06d11cbc405a12/tests/zds/extensions/smart_legend.txt#L30-L73
* https://github.com/zestedesavoir/Python-ZMarkdown/blob/f5fcec2db9eab20eaef17352414805551c10587a/tests/zds/extensions/smart_legend.html#L17-L66

PR #82, opened 2017-06-20
add table to rebber and caption

artragis:
I still need :

- [ ] rebber `gridTables`
- [ ] rebber `table` and `gridTable` captions.

This is linked to @pierre-24 work on tables inside the latex class.


vhf:
Please don't spend too much time on the rebber aspects of (grid)tables.

Priority goes to captions and other Markdown aspects. The LaTeX team hasn't decided how to handle tables yet: https://github.com/zestedesavoir/latex-template/issues/52

artragis:
So, if rebber is not what we need, everything is ok for now : captions are tested and working

packages/remark-captions/src/index.js
```diff
@@ -8,7 +8,10 @@ const legendBlock = {
 
 function plugin (opts) {
   if (opts) {
-    Object.keys(opts).map(key => legendBlock[key] = opts[key])
+    // not sure but I think I once saw a kind of "extend" function
+    Object.keys(opts).forEach(function (key) {
```
vhf: https://github.com/zestedesavoir/zmarkdown/blob/ed65ab653321d3ee3a81316e2e0d9fe729f0b116/packages/rebber/src/escaper.js#L3-L26 😉

packages/rebber/src/types/table.js
```diff
@@ -0,0 +1,31 @@
+/* Expose. */
+const one = require('../one')
+module.exports = table
+
+const defaultMacro = (ctx, node) => {
+  let j = 0
+  const parsed = []
+  node.children.map(n => parsed.push(one(ctx, n, j++, node)))
```
vhf: I think you're looking for `const parsed = node.children.map(n => one(ctx, n, j++, node))`

packages/rebber/src/types/table.js
```diff
@@ -0,0 +1,31 @@
...
+  const inner = parsed.join('')
+  const lengths = []
+  inner.split('\\hline\n').map(l => lengths.push(l.split('&').length))
```
vhf: Same here :)

packages/rebber/src/types/table.js
```diff
@@ -0,0 +1,31 @@
...
+  const cols = lengths.sort(cmp)[0]
+  let colHeader = '|'
+  let i = 0
+  for (;i < cols; i++) {
+    colHeader += 'c|'
+  }
```
vhf: What about `String.prototype.repeat`?

packages/rebber/src/types/tableRow.js
```diff
@@ -0,0 +1,17 @@
+/* Expose. */
+const one = require('../one')
+module.exports = tableRow
+
+const defaultMacro = (ctx, node) => {
+  let i = 0
+  const parsed = []
+  node.children.map(n => parsed.push(one(ctx, n, i++, node)))
```
vhf: What's the difference between this and `all(ctx, node)` ?

packages/remark-captions/README.md
```diff
@@ -8,3 +8,54 @@ Source: A little green man, with a saber larger than himself
...
+    <tbody>
+      <tr>
+        <td>bla</td>
+        <td>bla</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>figcapt1</figcaption>
+</figure>
+```
```
vhf: 😍

packages/zmarkdown/index.js
```diff
@@ -62,7 +62,7 @@ const zmdParser = (config) => {
     .use(remarkHeadingShifter, config.headingShifter)
     .use(remarkIframes, config.iframes)
     .use(remarkGridTables)
-    .use(remarkCaptions)
+    .use(remarkCaptions, config.captions)
```
vhf: 👍

packages/rebber/src/types/table.js
vhf: or  https://github.com/zestedesavoir/zmarkdown/pull/82#discussion_r123068777 ?

packages/remark-captions/src/index.js
artragis: thks, will get it tomorrow

packages/rebber/src/types/table.js
artragis: in fact I tried this but it didn't work. However in between i made mor changements so perhaps I can rollback to this.

packages/rebber/src/types/table.js
artragis: is that really exist?

packages/rebber/src/types/tableRow.js
artragis: all stringifies everything, here, I have an array of stringified

packages/remark-captions/README.md
artragis: yep, I thought of your pain to write all those readmes

packages/rebber/src/types/table.js
vhf: [String.prototype.repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) 💃

packages/rebber/src/types/table.js
vhf: I see. By the way if you need an index you won't reuse later on, remember that `map`'s callback is `(item, index, items) => {}`.

Which means you could drop the `j` in the parent scope and do something like: const parsed = node.children.map((n, j) => one(ctx, n, j, node))`

PR #86, opened 2017-06-21
begin ping plugin

AmarOk1412:
**DO NOT MERGE**

cf #37 

# TODO

- [ ] wait v24 (https://zestedesavoir.com)
- [ ] Use jest for tests.
- Configurable?


packages/remark-ping-zds/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    return value.indexOf('@', keep.index)
+  }
+  return -1
+}
+
+function plugin () {
+  function zdsMemberExists (username) {
+    //TODO https://github.com/zestedesavoir/zds-site/issues/4382
+    let isMember = false
+    const res = request('GET', `http://zestedesavoir.com/api/membres/?search=${username}`)
```
artragis: il faudrait que ça soit configurable, non?

packages/remark-ping-zds/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+  inlineTokenizer.locator = locator
+
+  const Parser = this.Parser
+
+  // Inject inlineTokenizer
+  const inlineTokenizers = Parser.prototype.inlineTokenizers
+  const inlineMethods = Parser.prototype.inlineMethods
+  inlineTokenizers.zping = inlineTokenizer
+  inlineMethods.splice(inlineMethods.indexOf('text'), 0, 'zping')
```
artragis: I think we should remove the `z` part for all `remarl_xxx` plugins.

packages/remark-ping-zds/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+      return eat(total)({
+        type: 'ping',
+        children: [{
+          type: 'text',
+          value: username
+        }],
+        data: {
+          hName: 'a',
+          hProperties: {
+            href: `/membres/voir/${username}/`,
```
artragis: Should be configurable.

packages/remark-ping-zds/src/index.js
AmarOk1412: Hmmm, I'm not sure actually. It's a very specific plugin for a specific case. If anyone want to use this plugin for another website, they must provide an endpoint which return a code 200 if the user exists.

vhf:
Thanks! That was the last important step!

PR #87, opened 2017-06-22
fix caption plugin conf issue

artragis:
I also added a start for #89


packages/zmarkdown/config.js
```diff
@@ -225,9 +225,16 @@ const config = {
...
   captions: {
-    table: 'Table:',
-    gridTable: 'Table:',
-    code: 'Code:',
+    external: {
+      table: 'Table:',
+      gridTable: 'Table:',
+      code: 'Code:'
+    },
+    inrernal: {
```
AmarOk1412: internal ;)

artragis: thx

PR #94, opened 2017-06-24
remark-captions: fix internal legends

vhf:
internal legends use the last legend


vhf:
Tiny bug to be fixed shortly:

```md
> > > Foo··
> > > Foo
> > Source: fc1
> >
> > Bar··
> > Bar··
> Source: fc2
>
> Baz
> Baz
Source: fc3
```

Here `fc2` is not parsed as legend because there's no `\nSource`

vhf:
Fixed.

PR #106, opened 2017-07-02
Fix dependencies

vhf:
Updated all dependencies; moved remark-parse from devdep to dep in zmarkdown because it's a remark-math peerdep.


vhf:
Ok, after investigating the issues we had with lock files and npm5 lately, I ended up finding that npm5 is simply not supported yet for use with Lerna

* https://github.com/lerna/lerna/issues/856
* https://github.com/lerna/lerna/issues/903

Issue #95, opened 2017-06-25
remark-captions: decide how img/Figure should work


vhf:
I'm not sure what's the desired behaviour for images with legends.

What I understand by reverse engineering zmd is:

```md
foo

![]()
Figure: 1 this is parsed as legend

baz

![]()
aFigure: 2 this text doesn't show up, img is wrapped in figure

![alt 2b](https://zestedesavoir.com/static/images/home-clem.4a2f792744c9.png)
this text doesn't show up, img is wrapped in figure, legend is "alt 2b"

foo
![]()
Figure: 3 this is displayed as text

![]()
Figure: 4 this is a legend, remainder of the paragraph goes into
the
legend
```

* I guess the main goal is to only parse a legend when the paragraph starts with an img.
* I'd vote for eliminating the behaviour illustrated by `2` and `4`
* Regarding `2b`, I would also drop support but keep the idea of using alt as legend if the paragraph only consists of an img. `2b` would be parsed as inline img with text following the image.

AmarOk1412:
+ I'd vote for eliminating the behaviour illustrated by `2` -> +1 (and +1 for eliminating `2b`)

+ Why do you want to eliminate `4`?

I agree with the rest.

artragis:
2b should be eliminated as it is not logical

3) should be "fixed" : as we force the Figure token, we should put the image inside a figure

vhf:
Ok, let's keep `4` and fix `3`.

PR #108, opened 2017-07-02
remark-captions: fix captions for images

AmarOk1412:
cf #95

### DIFF
--- a/packages/remark-captions/__tests__/index.js
+++ b/packages/remark-captions/__tests__/index.js
@@ -244,6 +244,11 @@ test('legend in paragraph', () => {
     Figure: 6 displayed as text
     ![]()
     Figure: 7 is a legend
+
+    ![]()
+    Figure: 8 is a legend.
+
+    Figure: 9 this is a text.
   `.replace(/·/g, ' '))
   expect(contents).toMatchSnapshot()
 })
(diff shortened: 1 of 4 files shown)


packages/remark-captions/__tests__/index.js
```diff
@@ -212,3 +212,38 @@ test('internal legend: two legends', () => {
...
+    aFigure: 2 this is displayed as text
+
+    ![alt 2b](https://zestedesavoir.com/static/images/home-clem.4a2f792744c9.png)
+    this is displayed as text
+
+    foo
+    ![]()
+    Figure: 3 this is a legend
+
+    ![]()
```
artragis: you forgot a similar case that is really usefull to handle : 

```markdown

![]()
Figure: the true figure

Figure: this one must stay as is.

File: packages/remark-captions/dist/index.js [hunk 2]
```diff
@@ -35,8 +35,8 @@ function internLegendVisitor(internalBlocks) {
     if (parent && parent.type === 'figure') return;
 
     // legend can only be in a paragraph
-    var lastP = getLast(node.children);
-    if (!lastP || lastP.type !== 'paragraph') return;
+    var lastP = node.children ? getLast(node.children) : parent;
+    if (!lastP || node.children && lastP.type !== 'paragraph' || !node.children && parent.type !== 'paragraph') return;
```
artragis: can you put a comment as a "4 conditional assertion with use of priority" is not the easyest thing to read.

packages/remark-captions/__tests__/index.js
AmarOk1412: @artragis updated!

artragis:
amha it's mergeable.

PR #112, opened 2017-07-06
iframe for rebber

artragis:
I just need to know why it is undefined inside tests

For failing tests in rebber I can't reproduce on my own PC as it uses symling to resolve submodules.


packages/remark-iframes/src/index.js
```diff
@@ -31,6 +31,18 @@ module.exports = function plugin (opts) {
...
+    if (provider.thumbnail && provider.thumbnail.format) {
+      thumbnail = provider.thumbnail.format
+      Object.keys(provider.thumbnail).filter(key => key !== 'format')
+        .forEach(function (key) {
+          thumbnail = thumbnail.replace(/`{${key}}`/g,
+            new RegExp(provider.thumbnail[key]).exec(url)[1])
+        })
+    }
+    return thumbnail
+  }
```
vhf: Replace logic isn't correct. You end up with a URL containing things like `/{id}/` which should have been replaced. This works:

```js
function computeThumbnail (provider, url) {
  let thumbnailURL = 'default image'
  if (provider.thumbnail && provider.thumbnail.format) {
    const thumbnailConfig = provider.thumbnail
    thumbnailURL = thumbnailConfig.format
    Object.keys(thumbnailConfig).filter(key => key !== 'format')
      .forEach((key) => {
        const search = new RegExp(`{${key}}`, 'g')
        const replace = new RegExp(thumbnailConfig[key]).exec(url)
        if (replace.length) thumbnailURL = thumbnailURL.replace(search, replace[1])
      })
  }
  return thumbnailURL
}
```

You can extract `computeThumbnail` and `computeFinalUrl` outside of the exported `plugin` function.

vhf:
Tests failing in rebber are too far away from their potential origin. Since it'll be eventually necessary anyway, you could start by testing if your changes to `remark-iframes` work properly (the thumbnail URL isn't properly generated for instance).

I'm pretty sure there's also an issue in the preprocess step, I'll comment on it later.

vhf:
```js
/* Stringify the given MDAST node. */
function toLaTeX (node, options, root) {
  preprocess(options, root)
  return one(options, node, undefined, undefined, root)
}
```

This runs the preprocess step on the `root`, then stringifies another node. I don't think that's correct. The preprocess step should run on the whole tree only once and before that. Otherwise the first parameter here will not be preprocessed, and what `one()` will get is a node that potentially isn't present in the root -- they are out of sync.

artragis:
I'll see to rebase and test the way to parse. But I don't get your point aboit the "they are out of sync". 

Can you provide a "more correct" snippet (even if incomplete) so that I understand?

packages/remark-iframes/src/index.js
artragis: I think it's really strange

`const replace = new RegExp(thumbnailConfig[key]).exec(url)` means "in url search the value that we found on the config with the given key". It will never find anything.

PR #170, opened 2017-09-08
remark-iframes: blocks instead of inline

vhf:
This breaks some zmd tests that rely on `remark-captions`.

@AmarOk1412 & @artragis could you help with captions?


artragis:
the first thing to do is to declare "iframe" as "external" in the remark-caption configuration.

then, I need to investigate why the caption is not well parsed

PR #195, opened 2018-02-09
Close #184: neutral blocks, custom block titles

### LINKED ISSUE #184: [Suggestion] Change the result of special blocks
Today, the special blocks are somewhere between [that kind of alert](http://docdock.netlify.com/shortcodes/alert/) and [this type of information](http://docdock.netlify.com/shortcodes/notice/), in the way that they do not have a header but the icon is imposed by the type of block.

The proposition is to add [this type of block](http://docdock.netlify.com/shortcodes/panel/), which will have these advantages, compared to existing:

- Ability to have "neutral" blocks.
- Ability to have a custom title.
- Ability to add a long text in a block without all the page being colored (really usefull with neutral blocks).
- Normally easy to add a type of block.

---------------------------

Topic's answers :

It would be a way to add the already mentionned blocks of type « Demonstration », « Theoreme », « Definition », numbered if possible.
The personalized blocks would have a limited choice of themes (at least : the existings + « neutral »), the title would be totally customisable.

Subject : https://zestedesavoir.com/forums/sujet/9492/suggestion-assouplir-la-gestion-des-blocs-speciaux/
*From zestedesavoir/zds-site#4716*
*Sent from Zeste de Savoir*

vhf:


artragis:
is it mergeable?

PR #201, opened 2018-02-19
Add compiler to remark-captions

artragis:


packages/remark-captions/__tests__/index.js
```diff
@@ -281,3 +290,75 @@ test('legend in paragraph', () => {
...
+    noop
+    
+    > foo
+    > bar
+    > baz
+    > qux
+    Source: **first**
+    b*a*r
+    Source: This is **the real** \`source\`
+  `
```
vhf: Please split these into a few separate cases. All the input syntax should be supported, for instance

    ```python
    print('bla')
    ```
    Code: figcapt1

returns `undefined figcapt1`

packages/remark-captions/__tests__/index.js
```diff
@@ -6,6 +6,7 @@ import remark2rehype from 'remark-rehype'
 
 import remarkCaptions from '../src/'
 import remarkGridTables from '../../remark-grid-tables/src'
+const remarkStringify = require('remark-stringify')
```
vhf: This should be an `import` placed above, see

https://github.com/artragis/zmarkdown/blob/b417f74a08e959cdb1a69e4075a20abab5ae66b3/packages/remark-captions/__tests__/index.js#L1-L6

PR #205, opened 2018-02-23
remark-abbr: add compiler

artragis:


packages/remark-abbr/__tests__/index.js
```diff
@@ -5,6 +5,7 @@ import stringify from 'rehype-stringify'
 import remark2rehype from 'remark-rehype'
 
 import remarkAbbr from '../src/'
+const remarkStringify = require('remark-stringify')
```
vhf: This should be an `import` above, l.1-5.

PR #217, opened 2018-03-02
add ico-after class in custom-blocks

artragis:


vhf:
I would rather change the CSS. I don't want think this is necessary, or am I missing something?

artragis:
don't know

PR #230, opened 2018-04-19
remark-mark: copy from remark-kbd

notapianokey:
Hello, 

please review my pull request for a new plugin. remark-mark parses custom Markdown syntax to handle spans of relevant text. It parses `!!marked text!!` into a new mdast inline node type `mark`, If you are using rehype the stringified HTML result will be `<mark>marked text</mark>`. [remark-kbd](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-kbd) already provides the necessary inline tokenizer so I copied most of its work.

This plugin is primarily intended for Inkdrop, a note taking tool that uses remark as its Markdown parser. I proposed a [plugin idea](https://forum.inkdrop.info/t/text-highlighting-plugin/425) on the forum that outlines the use case.

Thank you!
@notapianokey


vhf:
Nice, thanks @notapianokey !

packages/remark-mark/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "remark-mark",
+  "version": "1.0.0",
+  "repository": {
+    "url": "https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-mark",
+    "type": "git"
+  },
+  "author": "Victor Felder <victor@draft.li> (https://draft.li)",
```
artragis: I think you can add yourself to this one :)

notapianokey: Thanks :)

notapianokey:
af64513 fixes a problem where `!!ok foo!!!` would become `<mark>ok foo</mark>!` instead of `<mark>ok foo!</mark>`

packages/remark-mark/__tests__/index.js
```diff
@@ -0,0 +1,63 @@
...
+  This !!should 
+  work!! ok.
+
+  not!! marked!! text!!!
+
+  sdf !!!! df
+
+  sfdgs ! ! dfg !! dgsg ! qs
+
+  With two exclamation marks: \!!marked text!! you'll get !!marked text!!.
```
artragis: can you also add a test where the escape is at the end; e.g `!!not marked because of escape \!!`

artragis: good question @vhf ?

vhf: I don't think it is. No big deal though, I'll merge and fix it before releasing. I'll comment on this PR as soon as it's released.

vhf: Ok, I'm done. I couldn't find a good way of escaping end marks so let's go without them, to me it doesn't matter that much.

vhf:
@notapianokey The name of this package conflicts with https://www.npmjs.com/package/remark-mark , what would you suggest?

notapianokey:
Oh shoot. I only checked the list of plugins at [remarkjs](https://github.com/remarkjs/remark) and [zestedesavoir](https://github.com/zestedesavoir/zmarkdown) and completely missed there is already an implementation available on npm.

It also uses a different syntax. Following up on it I found some adoption for `==` at [remarkable](https://github.com/jonschlinkert/remarkable/tree/dev#syntax-extensions),
 [typora](http://support.typora.io/Markdown-Reference/), [ghost](https://help.ghost.org/article/4-markdown-guide) and a proposal on the [commonmark forum](https://talk.commonmark.org/t/highlighting-text-with-the-mark-element/840) dating back to 2014.

Although not an official spec, I would like to refactor this extension for the sake of interoperability and also use the existing syntax. Are you okay with that?

Regarding the npm package I will get in touch with the owner.

Sorry for the inconvenience!

PR #234, opened 2018-05-04
remark-custom-blocks: add compiler

artragis:


artragis:
@vhf the bug you found is fixed.

artragis:
does that mean I can merge?

PR #235, opened 2018-05-06
Stringify to markdown

arobase-che:
Explicit enough amha.

I just want to talk about 1 thing. I think that with the test : 

```markdown
+----- A -----+
|             |
+-------------+
|             |
+-------------+
```

Is a regression since now it's parsed as a table. See the [snapshot](https://github.com/zestedesavoir/zmarkdown/blob/d446425880f9337c8eddadc0e3fdb0b7c296dca6/packages/remark-grid-tables/__tests__/__snapshots__/index.js.snap#L47).

With love 💌

### DIFF
--- a/packages/remark-grid-tables/src/index.js
+++ b/packages/remark-grid-tables/src/index.js
@@ -448,6 +448,228 @@ function transformer (tree) {
   visit(tree, deleteWrapperBlock())
 }
 
+function createGrid (nbRows, nbCols) {
+  const grid = []
+
+  for (let i = 0; i < nbRows; i++) {
+    grid.push([])
+    for (let j = 0; j < nbCols; j++) {
+      grid[i].push({h: -1, w: -1, b: true, r: true})
+    }
+  }
+
+  return grid
+}
+
+function setWeight (grid, i, j, cols) {
+  /* To do it, we put enougth space to write the text.
+   * For multi-cell, we divid it among the cells. */
+  let W = Math.max(...Array.from(grid[i][j].value).map(x => x.length)) + 2
+
+  grid[i].forEach((_, c) => {
+    if (c < cols) { // To divid
+      const w = Math.ceil(W / (cols - c)) // cols - c will be 1 for the last cell
+      W -= w
+      grid[i][j + c].w = w
+    }
+  })
+}
+
+function setHeigth (grid, i, j, values) {
+  // To do it, we count the line. Extra length to cell with a pipe
+  // in the value of the last line, to not be confuse with a border.
+  grid[i][j].h = values.length
+  // Extra line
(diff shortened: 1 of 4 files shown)


packages/remark-grid-tables/src/index.js
```diff
@@ -448,6 +448,228 @@ function transformer (tree) {
...
+
+    // "End" line
+    // It's the last line of the cell. Actually the border.
+    line = row[0].b ? '+' : '|'
+
+    row.forEach((cell, j) => {
+      let c = ' '
+
+      if (cell.b) {
+        if (row[0].H) {
```
artragis: can you use more "obvious" name such as `isHeader`? idem for cell.b and cell.r etc.

arobase-che: Rename it to `isHeader`, `hasBottom` and `hasRigth` ? 

I'm out of coke/chocolate now, so yes but i'm out for an hour. How ? As a new commit or an amend one ?

Thanks for the review. It's really appreciated.

artragis: I think you should fixup the new commit to the old one.

AmarOk1412:
Ok (and hi btw, it's been a long time :( ). Here the answer (better than a PM)

# Empty grid

The min width of a column is 3. So, to get an empty table, you will write:

```
+---+
|   |
+---+
```

Not:
```
+--+
|  |
+--+
```

This is mainly for 2 reasons :

1. Empty table are not the main thing you want, and if there is a letter in a case, you will need at least 3 characters in the cell. (`| x |`)
2. Classic tables also needs at least 3 characters, so it's homogeneous.

# regression?

```
+----- A -----+
|             |
+-------------+
|             |
+-------------+
```

Is not a regression, but the behavior has changed yes. The thing is, now, we don't have to let a space between the table and the rest of the text.

```
This is an empty table:
+-------------+
|             |
+-------------+
```

Will give *This is an empty table:* followed by an empty table.

So,
```
+----- A -----+
|             |
+-------------+
|             |
+-------------+
```
will give 2 ugly lines followed by an empty table.

I think, this is logic, because we want this example as a valid grid-table followed by a text:

```
+---------+
| A       |
+---------+
Text at the end
```

So, this is the current behavior and I think it's the more logical thing to do, but if there is a consensus to change this, it's ok for me.

arobase-che:
Hi @AmarOk1412 \o :tada: 

About empty grid, now, i set the min width to 3.  (last commit)

About the weird behavior. For this PR, it doesn't matter and i don't know witch one is better.

AmarOk1412:
Indeed, nothing to do with this PR, just answering to your first post :)

artragis:
@vhf do you agree to merge this one?

PR #247, opened 2018-07-20
#241 - Wrap images

vhf:


artragis:
@vhf do you understand the cause of those failures?

PR #252, opened 2018-07-22
Do not ping in email address

### LINKED ISSUE #243: Ping and email adress
If one writes an email address, like `me@gmail.com`, the `gmail` user is pinged. A solution would be that the ping is activated only if it is preceded by a space or a newline (or nothing if it is the beginning of the md code).

```md
@test? → ping test
I call @test. → ping test
My address is test@test.com → no ping
```

pierre-24:
Closes #243 

(if I do something wrong, I apologize in advance)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
     "JSONStream": {
       "version": "1.3.3",
       "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.3.tgz",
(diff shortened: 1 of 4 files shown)


vhf:
Nice, thanks!

Merged in 8d31a6e3e9ac0c04ce31abe46efd60116a62feb9 , here's what I fixed:

- commit message
- tests
- reverted package-lock changes

pierre-24:
Just for curiosity, why did I change package-lock in the first place ?

vhf:
You might have been using a different npm version than I do. No big deal anyway, we don't need to use the same one. The rule is simple: only commit this file when you decided to modify it, for instance by installing/removing/updating a dependency.

PR #260, opened 2018-08-03
Add an alias for pinch smiley

nils-van-zuijlen:
Cf https://github.com/zestedesavoir/zds-site/pull/4960#issuecomment-410363280

### DIFF
--- a/packages/zmarkdown/config/remark.js
+++ b/packages/zmarkdown/config/remark.js
@@ -146,6 +146,7 @@ const remarkConfig = {
       'x(': '/static/smileys/pinch.png',
       '>_<': '/static/smileys/pinch.png',
       '&lt;_&gt;': '/static/smileys/pinch.png',
+      'X/': '/static/smileys/pinch.png',
       ':pirate:': '/static/smileys/pirate.png',
       ":'(": '/static/smileys/pleure.png',
       ':lol:': '/static/smileys/rire.gif',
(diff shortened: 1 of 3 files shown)


packages/zmarkdown/config/remark.js
```diff
@@ -146,6 +146,7 @@ const remarkConfig = {
       'x(': '/static/smileys/pinch.png',
       '>_<': '/static/smileys/pinch.png',
       '&lt;_&gt;': '/static/smileys/pinch.png',
```
vhf: Could we remove this now?

nils-van-zuijlen: Well... That's backward compatibility stuff.

vhf: Yeah, when we started this project we decided to drop support for a bunch of bad features and bugs used as features. That's one of them, we don't care about backward compatibility here.

nils-van-zuijlen: OK.
I think `x(` and `&lt;_&gt;` should be removed, but `>_<` should be kept because it was the official way of writing that smiley. `X/` is the new way of doing it.

vhf: To me x( makes sense. Let's only remove the weird html artifact that creeped in here because of an old bug in our editor.

vhf:
Merged and released, thanks! 620686e76ce5b1301b2c0cd77e4ad7d1bbdd245c

PR #261, opened 2018-08-03
Add smiley " :B "

A-312:
Cf. https://github.com/zestedesavoir/zds-site/pull/4951#issuecomment-410361589

### DIFF
--- a/packages/zmarkdown/config/remark.js
+++ b/packages/zmarkdown/config/remark.js
@@ -135,6 +135,7 @@ const remarkConfig = {
       ':colere:': '/static/smileys/angry.gif',
       'o_O': '/static/smileys/blink.gif',
       ';)': '/static/smileys/clin.png',
+      ':B': '/static/smileys/b.png',
       ':diable:': '/static/smileys/diable.png',
       ':D': '/static/smileys/heureux.png',
       '^^': '/static/smileys/hihi.png',


pierre-24:
It does now : https://github.com/zestedesavoir/zds-site/pull/4951/files 

... But I see the problem here: since zmarkdown is released more often than zds, it may cause issues ?

vhf:
Yep, zmarkdown doesn't follow zds releases. This smiley won't get merged before it actually exists, to make sure it exists I'll click on the link. If it shows a smiley, the smiley exists on zds, if it's a 404 then the smiley does not exist.

pierre-24:
It's a bit of a chicken-and-egg situation, but ok, I added a reminder for the next release ^^

vhf:
Not that much, there's a precise order in which to do this. If we released this here before zds, we would end up with something like this whenever somebody type `:B`

artragis:
here?

pierre-24:
I don't get why we still get a 404, even though that should be deployed in production ... o_O ?

artragis:
as far as I see it's only deployed on beta (will be added in v27.3)

pierre-24:
Ah, ok, I thought it was older than that ^^

A-312:
:B

A-312:
Ready

vhf:
I'll do this after https://github.com/zestedesavoir/zds-site/issues/5111

A-312:
@vhf #5111 is merged

Issue #257, opened 2018-07-27
zmarkdown: `latex` language not highlighted


artragis:
In pyZmarkdown, when we used

    ```latex
    \cos(n)^2 + \sin(n)^2 = 1
    ```

the code was coloured.

Now, we only support `tex` as language.

Is there a way to add `latex` as alias of `tex` in code highlighting?

vhf:
If we could get access to the `lowlight` instance, e.g. the one exposed here: https://github.com/wooorm/lowlight/blob/master/index.js#L5 , I guess we could create an [alias](https://highlightjs.readthedocs.io/en/latest/api.html?highlight=alias):

```js
low.registerLanguage('latex', require('highlight.js/lib/languages/tex'))
```

I'm not sure `rehype-highlight` exposes its lowlight instance though. One would need to investigate.

artragis:
I opened a pull request on rehype-highlight https://github.com/rehypejs/rehype-highlight/pull/6

PR #264, opened 2018-08-09
zmarkdown: highlight latex and keep console as plainText

artragis:
#257

### DIFF
--- a/packages/zmarkdown/__tests__/__snapshots__/new-html-suite.test.js.snap
+++ b/packages/zmarkdown/__tests__/__snapshots__/new-html-suite.test.js.snap
@@ -1,5 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`code highlight special cases does not highlight console 1`] = `
+"<div class=\\"hljs-code-div\\"><div class=\\"hljs-line-numbers\\"><span></span></div><pre><code class=\\"language-console\\">echo \\"Hello world\\"
+</code></pre></div>"
+`;
+
+exports[`code highlight special cases highlights latex as tex 1`] = `
+"<div class=\\"hljs-code-div\\"><div class=\\"hljs-line-numbers\\"><span></span><span></span><span></span><span></span><span></span></div><pre><code class=\\"hljs language-latex\\"><span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">usepackage</span><span class=\\"hljs-string\\">{inputenc}</span><span class=\\"hljs-string\\">[utf8]</span></span>
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">begin</span><span class=\\"hljs-string\\">{document}</span></span>
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">texttt</span><span class=\\"hljs-string\\">{code}</span></span>
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">end</span><span class=\\"hljs-string\\">{document}</span></span>
+\`\`\`
+</code></pre></div>"
+`;
+
 exports[`images become figures: does not apply when a caption is present 1`] = `
 "<figure><img src=\\"http://example.com\\" alt=\\"foo\\"><figcaption>foo</figcaption></figure>
 <figure><img src=\\"http://example.com\\"><figcaption>Caption</figcaption></figure>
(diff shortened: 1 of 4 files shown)


artragis:
I don't know why tests are failing here.

vhf:
Something changed in an upstream package we transitively depend on: https://github.com/syntax-tree/mdast-util-to-hast/commit/742a78036ff62895111e020034fb5225916d94a5

This fix makes `remark-parse` more compliant with the commonmark spec. We could either hold this dependency back or upgrade, I think it's better to upgrade. I'll ping you here as soon as it's fixed, you'll then have to rebase on master.

vhf:
I'll prepare this on a separate branch, I might wait on https://github.com/syntax-tree/mdast-util-to-hast/pull/22 before merging it.

vhf:
I decided to pin the dependency instead. You can now rebase on master.

artragis:
@vhf it's rebased.

packages/zmarkdown/__tests__/__snapshots__/new-html-suite.test.js.snap
```diff
@@ -1,5 +1,18 @@
...
+exports[`code highlight special cases does not highlight console 1`] = `
+"<div class=\\"hljs-code-div\\"><div class=\\"hljs-line-numbers\\"><span></span></div><pre><code class=\\"language-console\\">echo \\"Hello world\\"
+</code></pre></div>"
+`;
+
+exports[`code highlight special cases highlights latex as tex 1`] = `
+"<div class=\\"hljs-code-div\\"><div class=\\"hljs-line-numbers\\"><span></span><span></span><span></span><span></span></div><pre><code class=\\"hljs language-latex\\"><span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">\\\\</span></span>usepackage{inputenc}[utf8]
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">\\\\</span></span>begin{document}
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">\\\\</span></span>texttt{code}
+<span class=\\"hljs-tag\\">\\\\<span class=\\"hljs-name\\">\\\\</span></span>end{document}
```
vhf: From what I read here, only the `\` is highlighted, `end{document}` is not highlighted.

This seems wrong. TeX highlighting should at least detect `{}` and its content, and highlight this.

vhf: If you have issues reading snapshots, you don't have to use snapshots.

Or before using a snapshot, you could check the snapshot would be. Try `.toBe(null)` instead of `.toMatchSnapshot()`, it will usually tell you that you expected the output to be `null` but got `…` instead.

Read carefully what `…` is, and if you are 100% sure it is correct, use a snapshot.

artragis:
output was good, and in latex, but input was not good due to the fact I use template and did not escape `{` and `}`

artragis:
th'ats what I did in my last update and as you can see it's ok.

vhf:
It doesn't look right.

Why did you have to escape here: `\{document\}` but not here: `{inputenc}`?

artragis:
Ok, so,

```javascript
const input = dedent `\`\`\`tex
\\usepackage{inputenc}[utf8]
\\begin{document}
\\texttt{code}
\\end{document}
\`\`\``
```

I got the same result as when I ask for latex.

So two possibilities : 

- the output is really what we expect
- the input is not good, but I don't get where.

artragis:
It was the second option.

I had a look on how we test it on lowlight and rehype-highlight side and I just adapted the test with it.

It seems that I do not master the string format management in javascript. Can you explain it please?

When I did:
```javascript
const input = dedent `\`\`\`tex
\\usepackage{inputenc}[utf8]
\\begin{document}
\\texttt{code}
\\end{document}
\`\`\``
```

the result was not the same as

```javascript
const input = ['````latex',
      '\\usepackage{inputenc}[utf8]',
      '\\begin{document}',
      '\\texttt{code}',
      '\\end{document}',
      '```'].join('\n')
```

vhf:
Good job, that's much better :)

See how `dedent` behaves like [`String.raw`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw):

```js
> const dedent = require('dedent')
> dedent(`\\`)
'\\'
/* careful here: tagged template: */
> dedent`\\`
'\\\\'
> String.raw`\\`
'\\\\'
```

I'll rework the tests a bit to fix the formatting and I'll merge.

vhf:
Merged: bee1f544e4d1f0b5b5cc4a251c40e25f93346dfb

Issue #259, opened 2018-08-03
Run zmarkdown client-side


vhf:
The goal is to bundle zmarkdown for client-side use (live preview). Browserify, webpack, we don't have a preference, use whatever works for you.

Some of zmarkdown dependencies would be useless client-side, e.g. `rebber`, `rebber-*`, `remark-image-downloads`, … A way of not bundling them would be not to import them in the entrypoint that will be compiled, I guess.

We want to support a wide range of browsers, no support for IE < 11 though. This means transpiling using babel, including polyfills.

Heziode:
I will start working on it tomorrow.

vhf:
Yep, that was an easy one, I just pushed it because I wanted to fix the zmarkdown README together with it. It's a lot of work so I finally pushed the travis config only.

@Heziode Could you please explain what `check` does? It's mentioned in the readme but I don't understand the text that goes with it.

Heziode:
`check` is an optional function that take in argument a clone of `opts` passing to common. It can be used to allow the use of this plugin depending to the conf (if a some specific values are set to `true, etc.).

But, in hindsight, I think this function are useless…

vhf:
I'll do a PR fixing the doc and removing this

Heziode:
Ok, It's the best to do I think.

vhf:
Great work @Heziode !

PR #268, opened 2018-08-19
feat: Run zmd client side

Heziode:
This PR aims at providing a client-side bundle of zmarkdown.

Related to #259 

_**Note**: This is currently a **W**ork **I**n **P**rogress_

# Client Architecture

The architecture of the client is similar to `remark` or `Vue`. The *manager*, here `client/client.js`, exposes the variable `ZMarkdown`. This *manager* doesn't work alone because it does not know how to convert the input to the desired output. Example, if you want to convert markdown to html, the *manager* doesn't know how to do this. It need modules to know how to do this.

## Manager

You need to add modules with `ZMarkdown.use(obj)` to manage rendering.

To convert a markdown string you should use `ZMarkdown.render(str, moduleName = null, cb = null)`:
- **str**: `string`, string to convert
- **moduleName**: `string`, name of the module you want to use. This parameter can be **omitted only** if you define a default module with `ZMarkdown.setDefaultModule` (you can reset default type with `ZMarkdown.resetDefaultModule`)
- **cb**: `function(err, vfile)`, called when process is done

This function return a `Promise` if no callback specified.

`ZMarkdown` have also a `parse(moduleName)` function to get the MDAST tree and `getParser(moduleName)` to get the whole parser. This parameter can be **omitted only** if you define a default module with `ZMarkdown.setDefaultModule`.

## Module

The module is an object that should have the following properties:

- **name**: `string`, a string that define the name of the plugin. This is used to identify each plugins
- **render**: `function(input, cb): void`, a function called in `ZMarkdown.render`. It returns a `Promise` if no callback specified
- **parse**: `function(input): object`, get MDAST tree
- **getParser**: `function(): object`, get the whole of parser
- **initialize**: `function(config)`, configure the module with a custom configuration. You do not have to call this function if you want to use the default configuration 

# Tips

You can start a module with a base parser using `./common.js`. The module exports a function that can take two optionals parameters:
- **opts**: an object that can have: 
  - **remarkConfig**: `object` , your remark config (defaults to the configuration from `config/remark`)
  - **extraRemarkPlugins**: `array of objects`, remark plugins you want to add to the default parser (remark pipeline). The object should contain:
    - **obj**: `remark plugin`
    - **option**: optional, option of the plugin
    - **check**: `function(config)` optional, a function that returns a boolean to use this plugin depending to remark config passing in argument of `common`
- **processor**: `function(config)` (default `getHTMLProcessor` of `./common.js`), processor function used to configure the remark pipeline for you output

Module example:

```js
const common = require('../../common') /* zmarkdown common file */
const remarkToc = require('remark-toc')
const remark2rehype = require('remark-rehype')

const opts = {
  remarkConfig: null /* custom remark config. Null or omit to use default conf */,
  extraRemarkPlugins: [
    {
      obj: remarkToc,
      option: null /* remark plugin option. Null or omit to not configure it */,
      check: (config) => {
        /* 
         * This function allow to use this plugin only on certain cases. 
         * The config passed in argument are remark config.
         *
         * Note: This config are the config passed in opts when 
         * instancing common. Common make a clone of this conf, thereby
         * any update make after instancing common will not apear.
         * For example, the "noTypography" define in processor function
         * below not appear in this conf.
         */
        return true
      },
    },
  ],
}

const processor = (config) => {
  config.remarkConfig.noTypography = true

  return globalParser.zmdParser(config.remarkConfig, config.extraRemarkPlugins)
    .use(remark2rehype, config.remarkConfig.remark2rehype)
}

const globalParser = common(opts, processor)

export function render (input, cb) {
  return globalParser.render(input, cb)
}

export function parse (input) {
  return parser.parse(input)
}

export function getParser () {
  return parser
}

export const name = 'custom-html'
```

Modules need to be bundled for used on the client.

Here is a webpack conf of the previous example:

```js
const path = require('path')

const mode = process.env.NODE_ENV ? process.env.NODE_ENV : 'production'

module.exports = {
    mode,
    name: 'ZMarkdownCustomHTML', // name of process if you use parallel-webpack
    entry: ['./plugins/client/custom-html'], // path to your module
    output: {
      path: path.resolve(__dirname, 'dist'), // destination folder
      filename: 'custom-html.js', // file name
      library: 'ZMarkdownCustomHTML', // Name of the global constant for dom access
    },
    module: {
        rules: [
          {
            test: /\.js$/,
            exclude: /(node_modules|bower_components)/,
            use: {
              loader: 'babel-loader',
            },
          },
        ],
    },
}
```

# What was done

- [X] Modular client-side convert
- [X] ZesteDeSavoir HTML plugin
- [X] (Optional) ZesteDeSavoir Latex plugin
- [x] Investigate trouble with youtube video (remark-iframes): `Uncaught (in promise) TypeError: URLSearchParams is not a constructor`
- [x] Update reamde of zmarkdown package to add client description (present in this PR description)
# Todo


vhf:
@Heziode Any progress here?

Some guidance:

* The client-side zmarkdown entrypoint needs a different config from it's server-side counterpart
    * remove `remark-images-download`: the client shouldn't download images found in the markdown source
    * remove `rebber` and its deps: the client doesn't need to generate LaTeX source (at least in a first iteration ; LaTeX source preview might be a fun addition in the future but has very low priority since we don't need it for our use case)

Heziode:
No progress, i do not have a lot of time.
I will be able to devote myself entirely for this from Tuesday.

Heziode:
Instead of remove image download and latex transpiler, can we use an "plugin" architecture like Vue.js, remark and Cie ? Like that we can allow user (developer) to use what it want.

vhf:
Yes, all we currently have is plugins. You still need to remove image downloads because browsers don't have access to the filesystem.

Heziode:
Ok, I will do that.

Heziode:
@vhf Can you "pre-review" the code (for the structure) and tell me if you agree with this.
I will "clean code" and add tests later.

packages/zmarkdown/plugins/client/zhtml.js
```diff
@@ -0,0 +1,95 @@
...
+const rehypeHighlight = require('rehype-highlight')
+const rehypeHTMLBlocks = require('rehype-html-blocks')
+const rehypeKatex = require('rehype-katex')
+const rehypeLineNumbers = require('../../utils/rehype-line-numbers')
+const rehypePostfixFootnotes = require('rehype-postfix-footnote-anchors')
+const rehypeSlug = require('rehype-slug')
+
+const rehypeStringify = require('rehype-stringify')
+
+const jsFiddleAndInaFilter = node => {
```
vhf: Most of this file seems duplicated from the other one. What would you think of factoring the common parts out in a file that would define all of these utilities and exports them separately? (This file would also replace `zmdParser.js`)

```js
// …requires
// …definitions
module.exports = {
  foo,
  bar,
  baz
}
```

packages/zmarkdown/.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": ["env"],
```
vhf: Please add [`targets`](https://babeljs.io/docs/en/babel-preset-env#targets) in here

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,51 @@
+export const plugins = {}
+export let defaultType
+
+export function use (obj) {
+  if (!obj) {
+    throw new Error('This is not an object')
+  }
+
+  // Check the object structure
+  if (!('name' in obj)) {
```
vhf: Don't use `in` to check for the presence of a key, prefer `hasOwnProperty` (same for the rest of the file)

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error("missing 'name' in plugin")
+  } else if (typeof obj.name !== 'string') {
```
vhf: `else` is useless here (same for the rest of the file)

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+    throw new Error(`Unknown type: ${type}`)
+  }
+}
+
+export function resetDefaultType () {
+  defaultType = null
+}
+
+export function render (str, type = null, cb = null) {
```
vhf: Please make sure `cb` is used when passed as 2nd arguments and type isn't specified

packages/zmarkdown/client/index.js
Heziode: Can you please explain why use `hasOwnProperty` instead of `in` ?

packages/zmarkdown/plugins/client/zhtml.js
vhf: To me this would be a structural change, not a minor cleanup. Would you do it that way or what are your thoughts?

packages/zmarkdown/client/index.js
vhf: It's more precise and corresponds better to what we want to achieve here. We don't care about a method `obj.name()` (i.e. `YourObjectsPrototype.prototype.name`) for instance, what we're looking for is a property.

packages/zmarkdown/plugins/client/zhtml.js
Heziode: I have a little idea of ​​how to do that. I will do that tomorrow.

packages/zmarkdown/.babelrc
vhf: I'd say `> 0.5%, last 2 versions, Firefox ESR, not dead, ie >=11`

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,74 @@
...
+    case 'string':
+      if (type && !plugins.hasOwnProperty(type)) {
+        throw new Error(`Unknown type: ${type}`)
+      }
+      break
+    case 'function':
+      if (!cb) {
+        cb = type
+      } else if (typeof cb === 'function') {
+        throw new Error("Non deterministic use of cb due to type of 'type' and type of " +
```
vhf: Extra args are always allowed in JavaScript, we should silently ignore this case.

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+export function render (str, type = null, cb = null) {
+  if (plugins.length === 0) {
+    throw new Error('No plugins available.')
+  }
+
+  switch (typeof type) {
+    case 'string':
+      if (type && !plugins.hasOwnProperty(type)) {
+        throw new Error(`Unknown type: ${type}`)
```
vhf: Is `''` a known type?

Heziode:
@vhf, 

The feature is globally complete, but, I have some trouble with tests. I do not understand why it not pass. And now the "inspect" break the object.

packages/zmarkdown/client/index.js
```diff
@@ -0,0 +1,71 @@
+export const plugins = {}
+export let defaultType
+
+export function use (obj) {
+  if (!obj || (typeof obj !== 'object')) {
```
vhf: What is `!obj` for?

packages/zmarkdown/default.webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+  entry: ['babel-polyfill', './client/index'],
+  output: {
+    path: path.resolve(__dirname, 'dist'),
+    filename: 'zmarkdown.js',
+    library: 'ZMarkdown',
+  },
+  module: {
+    rules: [
+      {
+        test: /\.jsx?$/,
```
vhf: We can safely remove `x?` here

packages/zmarkdown/package-lock.json
```diff
@@ -0,0 +1,7733 @@
+{
```
vhf: Please delete this file for now

packages/zmarkdown/package.json
```diff
@@ -90,7 +96,9 @@
     "posttest": "npm run stop",
     "watch": "jest --watch",
     "server": "pm2 start -f server/index.js -i 3 --max-memory-restart 150M",
-    "stop": "pm2 kill"
+    "stop": "pm2 kill",
+    "watch:client": "webpack --watch --info-verbosity verbose --mode development --config default.webpack.config.js",
+    "watch:zhtml": "webpack --watch --info-verbosity verbose --mode development --config zhtml.webpack.config.js"
```
vhf: * What is the difference between these two?
* Would it make sense to add something for production builds?

packages/zmarkdown/public/css/md.css
```diff
@@ -0,0 +1,3 @@
+@charset "UTF-8";
```
vhf: Could you please document how to recreate/update this file and its sourcemap?

packages/zmarkdown/server.js
```diff
@@ -0,0 +1,41 @@
+const toVFile = require('to-vfile')
+const clone = require('clone')
+const rebberStringify = require('rebber/src')
+const zmarkdown = require('./common')
+
+const fromFile = (filepath) => toVFile.readSync(filepath)
+
+const remarkConfig = require('./config/remark')
+let zmd
```
vhf: It seems to me that nothing ever gets assigned to this variable?

packages/zmarkdown/server/handlers.js
```diff
@@ -27,7 +29,7 @@ const meters = endpoints.reduce((acc, endpoint) => {
   return acc
 }, {})
 
-// this object is used to memoize configured processors
+// this object is used to memoize confirenderFilegured processors
```
vhf: Typo

packages/zmarkdown/server/handlers.js
```diff
@@ -147,7 +149,11 @@ module.exports = function markdownHandlers (Raven) {
         }
       }
 
-      processors[key] = zmarkdown({remarkConfig: remark, rebberConfig: rebber}, target)
+      processors[key] = zmarkdown({remarkConfig: remark,
```
vhf: Please move the first key on the next line, with the same indentation as the other keys

packages/zmarkdown/zhtml.webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+  entry: ['./plugins/client/zhtml'],
+  output: {
+    path: path.resolve(__dirname, 'dist'),
+    filename: 'zmarkdown-zhtml.js',
+    library: 'ZMarkdownZHTML',
+  },
+  module: {
+    rules: [
+      {
+        test: /\.jsx?$/,
```
vhf: `x?`

packages/zmarkdown/zhtml.webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+        exclude: /(node_modules|bower_components)/,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            presets: ['env'],
```
vhf: Will the `.babelrc` be used that way or should we add `target` in here as well?

packages/zmarkdown/zmdParser.js
```diff
@@ -0,0 +1 @@
+// TODO: should be deleted
```
vhf: Forgot to delete this? 😉

yarn.lock
```diff
@@ -0,0 +1,4516 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
vhf: Please delete this file

.babelrc
```diff
@@ -5,5 +5,6 @@
     "development": {
       "sourceMaps": "inline"
     }
-  }
+  },
+  "targets": "> 0.5%, last 2 versions, Firefox ESR, not dead, ie >=11"
```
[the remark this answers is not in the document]
Heziode: Bad link ? I do not know where you want me to put it.

packages/zmarkdown/client/index.js
Heziode: Is to check if the object exist (not null not undefined, etc.)

packages/zmarkdown/package.json
Heziode: > What is the difference between these two?

One is just for the wrapper, the other one is for plugins zhtml. If we add zlatex, it we need another entry. But I will investigate the possibility to build all with only one file.

> Would it make sense to add something for production builds?

Of course, I am not adding this for now, but it is planned ;)

packages/zmarkdown/public/css/md.css
Heziode: This is juste a Copy/Paste of the file in `<project-root>/public`

Heziode: At the end in think I will update `<project-root>/public` and remove `<project-root>/packages/zmarkdown/public`

packages/zmarkdown/server.js
Heziode: I think this is the origin of my bug (test fails).

packages/zmarkdown/server/handlers.js
Heziode: :O I really wrote dirty code yesterday on the train…

packages/zmarkdown/client/index.js
Heziode: I have do that because `typeof obj` return `"object"`.

.babelrc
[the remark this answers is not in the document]
Heziode: Ok, my bad, I have understood.

packages/zmarkdown/zhtml.webpack.config.js
Heziode: > cacheIdentifier: Default is a string composed by the babel-core's version, the babel-loader's version, the contents of .babelrc file if it exists, and the value of the environment variable BABEL_ENV with a fallback to the NODE_ENV environment variable. This can be set to a custom value to force cache busting if the identifier changes.
[source](https://webpack.js.org/loaders/babel-loader/#usage)

So I can just let:
```js
use: {
    loader: 'babel-loader',
},
```

vhf:
Looking at the CI logs, it seems that zmarkdown tests are not using https://github.com/zestedesavoir/zmarkdown/blob/master/packages/zmarkdown/config/rebber.js anymore 😉

Heziode:
>Looking at the CI logs, it seems that zmarkdown tests are not using https://github.com/zestedesavoir/zmarkdown/blob/master/packages/zmarkdown/config/rebber.js anymore wink

I was already on it. Well, we get closer to the goal, but it's not even that :sweat_smile:

Heziode:
The order of plugins used by remark are important ?

vhf:
Not for all of them. In general I'd like the order to be preserved, it makes debugging much easier.

Heziode:
Ok, I will try to get the same order, but, it can be hard in "agnostic" archictecture.

packages/zmarkdown/zhtml.webpack.config.js
vhf: Same with a production build. Current target gives `92467B`, ie 6 target: `92467B`, no target `92467B`

vhf:
> Ok, I will try to get the same order, but, it can be hard in "agnostic" archictecture.

Could you please explain how `client/client.js` is supposed to be used?

What is a plugin type?

Heziode:
@vhf I have updated the PR description. Tell me if it answer to your question.

packages/zmarkdown/public/css/md.css
vhf: All we need is a symlink here. Not sure what the question means. :)

Heziode: Nope ^^

vhf:
I fixed what I could: https://gist.github.com/vhf/92b10f611cfc39a367fe64ae43595ab7/revisions

The `[???]` indicate the parts I don't understand.

* Could you please provide an example?
* What is a plugin? How do they differ from a remark or rehype plugin, e.g. `remark-captions`?
* I'm confused by *plugin types*. If the plugin type is a plugin name, could we name it `name`?

Heziode:
> What is a plugin? How do they differ from a remark or rehype plugin, e.g. remark-captions?
My apologize, I have used `plugin` word for two differents things.

I have added comment in you gist.

vhf:
Good progress. Could you please remove codacy (we don't use it in this project) and update the README?

After that I'll try this PR again locally (building it using the instructions you'll provide), do a last round of review and we'll be good to go.

Heziode:
It is done.

Heziode:
@vhf if you don't use codacy for this project, can you disable it ? It fails the PR.

artragis:
oups, my bad.

packages/zmarkdown/common.js
```diff
@@ -127,34 +120,41 @@ const zmdParser = (config, target) => {
         vfile.fail(`Markdown AST too complex: tree depth > ${config.maxNesting}`)
       }
     })
-}
 
-function getLatexProcessor (remarkConfig, rebberConfig, target) {
-  remarkConfig.noTypography = true
+  if (extraRemarkPlugins && extraRemarkPlugins.length > 0) {
```
vhf: I think you could remove this condition if you do `const zmdParser = (config, extraRemarkPlugins = []) => {` on l.70.

packages/zmarkdown/common.js
```diff
@@ -127,34 +120,41 @@ const zmdParser = (config, target) => {
...
+    for (const record of extraRemarkPlugins) {
+      if (!record.check || record.check(config)) {
+        if (record.option) {
```
vhf: In the README you wrote that passing `null` as option means no configuration for the plugin. Is this really what we want? What if a plugin has a default config and passing `null` disables it? In this case we couldn't pass `null`.

I think there is some confusion between the semantics of `null` and `undefined`. A missing config would be `undefined`, it would use a default value if any. Explicitly passing `null` should still be possible.

packages/zmarkdown/common.js
```diff
@@ -206,26 +206,27 @@ function getDepth (node) {
...
-      remarkConfig is missing!`)
+  if (!opts) {
+    opts = {}
   }
-  if (!opts.rebberConfig || !Object.keys(rebberConfig).length) {
-    throw new Error(dedent`
-      This module expects to be called with ({remarkConfig, rebberConfig}, target = 'html'),
-      rebberConfig is missing!`)
+
+  if (!opts.remarkConfig || (opts.remarkConfig === null) || !Object.keys(remarkConfig).length) {
```
vhf: If `opts.remarkConfig === null`, then `!opts.remarkConfig === true`. ;)

packages/zmarkdown/config/remark.js
```diff
@@ -29,6 +29,7 @@ const remarkConfig = {
     blocks: [],
   },
 
+  canUseTextr: true,
```
vhf: I'd prefer something like `useTextr` or `enableTextr` or (with the value as `false`) `disableTextr`. It's not that this option let's us decide later whether we want to use textr or not, this option explicitly enables it.

packages/zmarkdown/server.js
```diff
@@ -0,0 +1,60 @@
...
+  config.remarkConfig.noTypography = true
+
+  return zmd.zmdParser(config.remarkConfig, config.extraRemarkPlugins)
+    .use(rebberStringify, config.rebberConfig)
+}
+
+function rendererFactory (config, target) {
+  if (target === 'html') {
+    return zmd.rendererFactory(config)
+  } else if (target === 'latex') {
```
vhf: It seems to me that `else` is useless here?

Heziode:
I do that asap.

vhf:
Apparently there's no way for me to remove @codacy. This thing is really broken. Nothing I can do from this repo settings panel, clicking on the failed @codacy check redirects me to a login page where this service wants me to give it access to my github profile, which I'll never allow.

artragis:
Logicaly, on zmarkdown this should be fine as no other member would have set it up. The currrent "failure" was due to the fact I, one day, added zmarkdown in my own codacy dashboard. I removed it, so on next commit my "codacy warning" won't show up. 
I clearly agree : codacy is broken.

packages/zmarkdown/common.js
vhf: What I meant was: if `x` is `null`, then `!x` is always `true`.

A condition like `if (x === false || !x)` is redundant.

packages/zmarkdown/common.js
vhf: I still think this condition is incorrect. I think you are confusing `null` with `undefined`.

Imagine the following plugin:

```
(config) => {
  if (typeof config === 'undefined') config = defaultConfig // this is basically (config = defaultConfig) => {}
  doSomething(config)
}
```

How do I pass `null` to `doSomething`? (`if (record.option)` will be `false` when `record.option === null`, which is why this condition is not correct.)

packages/zmarkdown/common.js
```diff
@@ -121,14 +121,12 @@ const zmdParser = (config, extraRemarkPlugins) => {
...
-    for (const record of extraRemarkPlugins) {
-      if (!record.check || record.check(config)) {
-        if (record.option) {
-          mdProcessor.use(record.obj, record.option)
-        } else {
-          mdProcessor.use(record.obj)
-        }
+  for (const record of extraRemarkPlugins) {
+    if (!record.check || record.check(config)) {
+      if (record.option) {
```
vhf: See https://github.com/zestedesavoir/zmarkdown/pull/268#discussion_r222430420

packages/zmarkdown/common.js
vhf: In other words: the config should un*defined* unless you *define* some config.

Defining the config as `null` means that `null` is the config, it is defined as `null` and therefore it is not undefined.

vhf: There are plenty of other cases.

`if (config) useConfig(config)` doesn't let you use `null` or `false` or `0` or `''` as config. All of these values could be used by plugins authors.

packages/zmarkdown/common.js
Heziode: It is changed.

packages/zmarkdown/common.js
Heziode: Indeed, I have do the change.

packages/zmarkdown/common.js
Heziode: It is changed.

packages/zmarkdown/README.md
```diff
@@ -202,7 +202,7 @@ const opts = {
   extraRemarkPlugins: [
     {
       obj: remarkToc,
-      option: null /* remark plugin option. Null or omit to not configure it */,
+      option: undefined /* remark plugin option. undefined or omit to not configure it */,
```
vhf: Thanks! Now the README needs to be updated accordingly, that's where I spotted this `null` idea

packages/zmarkdown/public/index.html
```diff
@@ -0,0 +1,201 @@
...
+  <div id="center-top" class="split split-vertical">
+    <div class="result render zmarkdown"></div>
+  </div>
+  <div id="center-bottom" class="split split-vertical"><pre class="bottom-style"><code></code></pre></div>
+</div>
+<div id="right" class="split split-horizontal">
+  <pre class="result render latex-result">
+  </pre>
+</div>
+<script src="../dist/zmarkdown.js"></script>
```
vhf: This will be an issue to deploy to github pages because this file will be at the root of the branch.

I suggest to delete the current root `public` symlink and creating a symlink from here `./dist -> ../dist`. In the travis deploy script I'll then copy the content before pushing.

vhf: Here would be the travis config:

```yml
git:
  depth: 10

language: node_js

cache:
  directories:
    - node_modules

node_js:
  - '8'
  - '10'

env:
  - DEST=/tmp
  global:
    secure: V3Wlrzc7mpyVxNxeBPMkfyowycurM2OmkTTc3aeUkRjrdltRNfknudhC2OELSpsfMvUTLLSPW/oMfv+jxgVHOydVIsBbN+lB0jKnm/T+ABfUeDw+tpEEvKvKLtD/uxDB3rC83uD9ociE82PqdGYtBOJX4Sug5bdoq8n02DlsGd1dnsr1dIdIIX0LHp4OkfRF6jv0EyXwSHK7Kk6s/miPtuChHrhJQwsjJBu5YPdSEu+f0mzilEO5NUiv4hT1IWjfBravP6uTyWIUcwtfY1rgmjZy3d+MaCRKbLi4hrzrIa/VL/rX9fg8vN6Jgnt1+kJtPB7Y0yH2tcl0tpeRWW2ppBZOUU2rJ5OWA8TKltZZW7zjQgrdrwNns0+8ifZ7AI/Jdfvfj0NjW/SqsxchLQL2uN0FsBkoUV6Ly4cwZ/B5EIFTdi0qnjcYBdHoXE21j81clk0rKdgiof/lGOL4+bA59O7V88AEOmgGshRF1AUbe1SwBQ6l8+w44ZDg7I9OQfNSf/W07NbT4cwSF+xp4nc0Rw0EzHnfHLxdYv0Dj+1GxAQdtJ22wmN5JFKlENsfslC6bHT2ZN2V5U0K2VqCBdh0iNdV/uV9tdy3GCSnO3W2XuBKtDbo+wrLf8rjIubWwupej8JxR+pPJ10KNJN5rhe/pqIRujr5FxUZJG7mll9w9Ws=

before_install:
  - npm install -g pm2

install:
  - npm install
  - npm run bootstrap

before_script:
  - pm2 ping

after_success:
  - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
  - rm ./packages/zmarkdown/public/js
  - cp ./packages/zmarkdown/dist/* ./packages/zmarkdown/public/js/

deploy:
  provider: pages
  skip-cleanup: true
  github-token: "$GITHUB_TOKEN"
  keep-history: false
  target-branch: gh-pages
  local-dir: ./packages/zmarkdown/public
  on:
    branch: master
```

Heziode: If I well understood what was asked, it's good now.

PR #272, opened 2018-09-01
remark-emoticons: fix leading spaces eating bug

vhf:
A trimming issue led to the space in `*foo* :)` to be incorrectly eaten.


vhf:
- rebber-plugins@1.4.8
 - remark-emoticons@2.1.0
 - zmarkdown@5.9.3

PR #286, opened 2018-09-10
rebber/zmarkdown : enhance table rendering in latex

### LINKED ISSUE #282: [rebber] Get tables with almost the same behavior as in HTML
As requested by artragis in https://github.com/zestedesavoir/latex-template/issues/93, here is how to get table that behave almost like their HTML counterparts.

```
C1 | C2 | C3
----|------|-----
short| miiiidle | looooooooooong
``` 

Should be rendered

```latex
\begin{longtabu} spread 0pt {|X[-1]|X[-1]|X[-1]|} \hline
 C1& C2 & C3 \\
\hline
short& miiiidle&looooooooooong\\
\hline
\end{longtabu}
```

notice `spread 0pt`, which gives table a width smaller than the page width if taking that much space is not required.

screenshot from 2018-09-06 21-05-47

Then `X[-1]` is a (very) special construct that seems to allow columns to takes as much space as they need (but with respect to other columns) but also insert line return if the column is too large. Therefore, this is what we need.

screenshot from 2018-09-06 21-06-24

If you still prefer that the text is centred (it was the previous behaviour), it is still possible with `X[-1c]`. But personally, I would not recommend it. 

--- 

And if you want to get even closer to the website experience, consider

```latex
\begin{longtabu} spread 0pt {|X[-1]|X[-1]|X[-1]|}
\hline
\rowfont[c]{\bfseries} % first line is with centred-align cells in bold
 C1& C2 & C3 \\
\hline
\rowfont[l]{} % other lines with left-aligned cells. No need to repeat that, once is enough
short & miiiidle & looooooooooong \\
\hline
\end{longtabu}
```

Which gives:

screenshot from 2018-09-06 21-27-13

(there is also an option for background color, but do we want that?)

artragis:
fix #282

### DIFF
--- a/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
+++ b/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
@@ -217,8 +217,10 @@ quote
 
 
 
-\\\\begin{longtabu}{|c|c|} \\\\hline
+\\\\begin{longtabu} spread 0pt {|X[-1]|X[-1]|} \\\\hline
+\\\\rowfont[c]{\\\\bfseries}
 Nom & Age \\\\\\\\ \\\\hline
+\\\\rowfont[l]{}
 Fred & 39 \\\\\\\\ \\\\hline
 Sam & 38 \\\\\\\\ \\\\hline
 Alice & 35 \\\\\\\\ \\\\hline
@@ -226,8 +228,10 @@ Mathilde & 35 \\\\\\\\ \\\\hline
 \\\\end{longtabu}
 
 
-\\\\begin{longtabu}{|c|c|} \\\\hline
+\\\\begin{longtabu} spread 0pt {|X[-1]|X[-1]|} \\\\hline
+\\\\rowfont[c]{\\\\bfseries}
 Nom & Age \\\\\\\\ \\\\hline
+\\\\rowfont[l]{}
 Fred & 39 \\\\\\\\ \\\\hline
 Sam & 38 \\\\\\\\ \\\\hline
 Alice & 35 \\\\\\\\ \\\\hline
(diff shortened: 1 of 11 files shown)


artragis:
@pierre-24 can you check if the generated latex is what you expect?

packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
```diff
@@ -255,8 +259,11 @@ a paragraph\\\\footnote[l9c12o138]{\\\\label{footnote:l9c12o138} footnoteRawPar
 
 exports[`gridTable 1`] = `
 "\\\\begin{longtabu}{|p{\\\\linewidth / 3}|p{\\\\linewidth / 3}|p{\\\\linewidth / 3}|} \\\\hline
```
pierre-24: Not that I'm against it, but what is the technical reason for maintaining this construct while `X[-1]` should do it ? (if you want three columns that have the same size but do not want to cares about the number of columns, you can also consider `X[1]`, which will give column of equal sizes no matter the number).

pierre-24:
Sorry, totally forget about that one. It looks good to me (modulo my comment above) :)

packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
artragis: In fact as your issue was only for normal tables I thought you wanted the gridtable behaviour to be kept. that's why you have the old syntax here.

artragis: Did you manage to check this behaviour?

pierre-24:
Oh, yeah sorry, not the best time this week, I'll try to catch up on the
other PR this weekend. Anyway, I did try and long story short, keep it like
this for now, it will be easier (multicolumn is NOT designed to work with
tabu, so what you do is probably the best option). Sorry for being long,
looks good to me :)

artragis:
ok, after @pierre-24, do you think we are ok @vhf ?

packages/rebber/src/types/tableRow.js
```diff
@@ -9,8 +9,18 @@ const defaultMacro = (ctx, node) => {
...
 
+const defaultFirstLineRowFont = `\\rowfont[c]{\\bfseries}`
+const defaultOtherLineRowFont = `\\rowfont[l]{}`
+
 /* Stringify a tableRow `node`. */
-function tableRow (ctx, node) {
+function tableRow (ctx, node, index) {
   const macro = ctx.tableRow || defaultMacro
-  return macro(ctx, node)
+  const firstLineRowFont = ctx.firstLineRowFont || defaultFirstLineRowFont
```
vhf: Is the first line actually a header?

artragis: good question, in fact this was the old behaviour and i kept it here because I still don't know how to resolve the true header

artragis:
@vhf I updated the documentation, waiting for your review.

For the header/firstLine, as I said, it was an existing code, and I think it is not in the scope of this PR.

vhf:
I'll fix the README and merge manually.

artragis:
thanks @vhf

PR #288, opened 2018-10-08
fix(smiley): fix >_< smiley at the beginning of the paragraph

### LINKED ISSUE #281: remark-emoticon/zmarkdown can't start a paragraph by smiley >_<
When we write:

```
>_< hello
```

we get

```
<blockquote>
<p>_&#x3C; hello</p>
</blockquote>
```

What is expected is  the same result as 

```
X/ hello
```

i.e

```
<p><img src="/static/smileys/pinch.png" alt="X/" class="smiley"> hello</p>
```

Heziode:
`>_<` smiley have now an alternate way to display it: `:>_<:`.
This allow to start a paragraph with this smiley and to not collide with:
- CommonMark spec: [example 192](https://spec.commonmark.org/0.28/#example-192)
- GFM spec: [Example 200](https://github.github.com/gfm/#example-200)

Fix #281

### DIFF
--- a/packages/zmarkdown/__tests__/__snapshots__/new-html-suite.test.js.snap
+++ b/packages/zmarkdown/__tests__/__snapshots__/new-html-suite.test.js.snap
@@ -34,7 +34,17 @@ exports[`pedantic mode disabled unordered lists markers 1`] = `
 </ul>"
 `;
 
+exports[`smileys translates :>_<: 1`] = `"<p>This is funny <img src=\\"/static/smileys/pinch.png\\" alt=\\":>_<:\\" class=\\"smiley\\"></p>"`;
+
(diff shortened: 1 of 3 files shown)


vhf:
IMO this PR doesn't solve the issue. Merging it wouldn't let you start a paragraph with `>_<` and get a smiley, and the suggested syntax isn't better than recommending the use of other alternatives such as `x(`.

Heziode:
In any case this problem can not be corrected unless the spec of CommonMark and GFM changes.

What do you propose ?

vhf:
Since block tokenizers and inline tokenizers have separate priority lists, I would go with wontfix or remove this smiley.

Heziode:
Ok do that.

Issue #295, opened 2018-11-08
zmarkdown/remark-math: katex 0.10.x is out


artragis:
as said in the title, remark 0.10.1 is out. This version brings important functionalities that zds folks need for their tutorials.

vhf:
We are not maintaining remark-math or am I missing something?

artragis:
in fact, I need to propose a PR to rehype-katex & then to zmarkdown to update the dependency.

artragis:
rehype-katex has a new version https://github.com/Rokt33r/remark-math/commit/0c4ab5022e30624db8739fc038c5d983a3c2ebd4

PR #297, opened 2018-11-14
zmarkdown: upgrade rehype-katex

artragis:
use new version of rehype-katex

#295

### DIFF
--- a/packages/zmarkdown/package.json
+++ b/packages/zmarkdown/package.json
@@ -28,7 +28,7 @@
     "rehype-footnotes-title": "^0.0.27",
     "rehype-highlight": "^2.2.0",
     "rehype-html-blocks": "^0.0.28",
-    "rehype-katex": "^1.0.2",
+    "rehype-katex": "^1.2.0",
     "rehype-postfix-footnote-anchors": "^1.0.3",
     "rehype-slug": "^2.0.1",
     "rehype-stringify": "^4.0.0",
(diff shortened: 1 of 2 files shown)


artragis:
@vhf, je ne comprends pas le test qui plante (notons que j'ai bien compris qu'il plante car viewBox n'est pas dans le string ni preserveAspectRatio) mais je ne comprends pas pourquoi ces éléments doivent être là.

vhf:
@artragis Yep! It's related to this: https://github.com/rehypejs/rehype/issues/2

Katex renders (used to render?) formulas as SVG with a `viewBox` attribute (and other camelCase attributes). Rehype doesn't handle these properly: they become e.g. `view-box`.

I wrote a hack based on the complete list of possible SVG attributes: https://github.com/zestedesavoir/zmarkdown/blob/1702c551a35e6490d01617a8258ebd989a0b8e53/packages/zmarkdown/utils/svg-hack.js

Assertions such as these: https://github.com/zestedesavoir/zmarkdown/blob/201c249a128291cd8dc545001133467b4aaccbc9/packages/zmarkdown/__tests__/server.tests.js#L118-L121

are making sure that `viewBox` isn't replaced by `view-box`, etc.

These tests could be failing for two reasons: either Katex changed how they render things and it's not an SVG with a `viewBox` attr anymore or Katex rendering is broken in our project. 🤽‍♂️ 🐃

vhf:
Yep, this comes from rehype@1.2.0 vs 1.1.1, cf. https://gist.github.com/vhf/62ddb4e6967b8ca1a05d2c18572c20cd

Here is one of these SVG parts of formula, probably a symbol that got replaced by something that doesn't need SVG:

```svg
<svg width="400em" height="0.2em" viewBox="0 0 400000 200" preserveAspectRatio="xMinYMin slice">
<path d="M0 80H400000 v40H0z M0 80H400000 v40H0z">
</path>
</svg>
```

vhf:
#301

PR #298, opened 2018-11-16
fix(live-demo): Fix external dependencies

Heziode:
Again… this is due to Split.js lib (and unpkg redirect).


vhf:
Don't worry about the tests, they will be fixed soon.

vhf:
Fixed.

PR #315, opened 2019-02-21
mdast-util-split: start of work

artragis:
# Context

this PR aims to add a new mdast-util project to prepare some work on zmarkdown that will try to import a flat markdown text to a fully hierarchized tutorial/article.

For that I need an util to traverse all headings and extract "part" from what follow those headings.

# What is done 

- [x] the basic documentation
- [x] the canonical case (presented in doc & in tests)

# What remains before merging

- [x] test of all configuration (for example I'm not sure my first commit properly handles the `conclusion` parameter)
- [x] test edge cases : what if a sub header is inside a quote? At the beginning I was using `unist-util-find` but as it finds everything in depth I changed the logic to a custom function that I haven't fully tested yet
- [x] test edge cases : what if there is no header? don't forget to mention `mdast-util-normalize`
- [x] implement the "sub header parsing" create a more splitted document (example with
part > subpart > chapter)

# Why doing such a PR?

- I'd like to have some comment about the functionality
- Is it a good way to perform what I want?
- Is my documentation good enough?
- Have you ideas of edge cases?
- Have you comment on my code style?
- Hound please help me improve code quality.


vhf:
Nice! I'll do a few rounds of reviews, the first one will only be obvious things, later on I might PR against your branch to fix trickier things.

I'm not sure about the naming of this package. `mdast-util-split` sounds very generic while this package seems to solve a specific thing.

packages/mdast-util-split/README.md
```diff
@@ -0,0 +1,490 @@
+# mdast-util-split [![Build Status][build-badge]][build-status] [![Coverage Status][coverage-badge]][coverage-status]
+
+
+**rebber** is a LaTeX stringifier for [remark][]
```
vhf:
```suggestion
```

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,124 @@
+module.exports = split
+import visit from 'unist-util-visit'
+
+function split (tree, {splitDepth = 1,
```
vhf: `splitDepth` is not used.

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+  introductionAsProperty = true, conclusionAsProperty = false}) {
+  if (splitDepth < 1) {
```
vhf: Isn't this assertion redundant with the default value set in the constructor?

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+    throw new Error('splitDepth must be greater than 1')
+  }
+  const firstHeading = find(tree, {type: 'heading', depth: 1})
```
vhf: `1` is a magic number

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+        return
+      }
+      splittedPart.conclusion = {
+        type: 'root',
+        children: splittedPart.children.splice(lastIndex, splittedPart.children.length - lastIndex),
+      }
+    })
+  }
+  extractIntroductions () {
+    const self = this
```
vhf: No need for this trick

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+      return tree.children[i]
+    }
+  }
+  return null
+}
+
+class Splitter {
+  constructor (depth = 1) {
+    this.lastIndex = -1
+    this.treeBefore = {
```
vhf: What is `treeBefore`?

packages/mdast-util-split/src/index.js
artragis: not yet implemented

packages/mdast-util-split/src/index.js
artragis: one could try to passe a negative integer. I don't know if in JS we make assertions when the boundaries are documented.

vhf: I wouldn't since passing `true` or `'foo'` wouldn't throw here, yet they aren't positive integers >= 1.

packages/mdast-util-split/src/index.js
artragis: Now it's implemented, but I'm not happy at all with my namings, I always mix "trees" and "children" and it's quite complex when splitDepth is greater than 1.

packages/mdast-util-split/src/index.js
artragis: Removed.

packages/mdast-util-split/src/index.js
artragis: Removed, was something I created in a first attempt to extract introduction.

packages/mdast-util-split/src/index.js
artragis: How can I do it then? if I use this in the lambda, this is nill.

packages/mdast-util-split/src/index.js
artragis: Fixed, nice spot.

packages/mdast-util-split/src/index.js
vhf: No, `this` will be the object, not `undefined`.

packages/mdast-util-split/src/index.js
vhf: It's confusing I agree. I think `this.subTrees` actually refers to new root trees, not subtrees?

Try drawing a simple schema of the output for a simple md doc at various `splitDepth`, it should make it easier to see what is a root tree, what is a subtree (i.e. a subpart of mdast root tree)?

vhf:
```md
the introduction

# chapter 1

a paragraph

> a quote to *ensure this is parsed*

## section 1.1

content 1.1

## subsection 1.1.1

content 1.1.1

## section 1.2

content 1.2

# chapter 2

## subsection 2.1.1

content 2.1.1

# the conclusion

paragraph
```

For instance what is expected here with splitDepth 1 and splitDepth 2?

My guess for splitDepth 1:

```js
{
  introduction: {}, // root tree from doc root to `chapter 1` (not included)
  trees: [
      {}, // root tree from `chapter 1` to `chapter 2` (not included)
      {}, // root tree from `chapter 2` to `conclusion` (not included)
  ],
  conclusion: {}, // root tree from `chapter 2` (not included) to document end
}
```

artragis: For this one, it is what I want.

artragis: In fact, not, the `conclusion` must be from "conclusion" (not included)

vhf: I guess this is what you mean:
```
{
  introduction: {type: root, children: [/* from doc root to `chapter 1` (not included) */]},
  trees: [
      {type: root, children: [/* from `chapter 1` to `chapter 2` (not included) */]},
      {type: root, children: [/* from `chapter 2` to `conclusion` (not included) */]},
  ],
  conclusion: {type: root, children: [/* from `conclusion` (not included) to document end */]},
}
```

What about splitDepth 2 for this document?

packages/mdast-util-split/README.md
```diff
@@ -0,0 +1,496 @@
...
+#### `options.splitDepth = 1`
+
+An integer greater or equal to 1 determining the max heading depth at which the tree is split.
+
+#### `options.introductionAsProperty = true`
+
+A boolean driving the way we want to process introduction :
+- if `true`, all MDAST elements between the split header and the first sub header will be extracted and put in the `introduction` property of the result
+property of the returned object
+- if `false` they will just be added as MDAST elements of the `children` property.
```
vhf: I don't understand this line, could you please clarify?

packages/mdast-util-split/README.md
```diff
@@ -0,0 +1,496 @@
...
+
+#### `options.conclusionAsProperty`
```
vhf:
```suggestion
#### `options.conclusionAsProperty = false`
```

packages/mdast-util-split/README.md
```diff
@@ -0,0 +1,496 @@
...
+
+A boolean driving the way we want to process the last sub header if there are more than one.
+- if `true`, the last sub header is removed from tree and all following elements are stored in the `conclusion` property of the result
+- if `false` (default), nothing is changed.
```
vhf:
```suggestion
- if `false`, nothing is changed.
```

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+class Splitter {
+  constructor (depth = 1) {
+    this.lastIndex = -1
+    this.subTrees = []
+    this.depth = depth
+    this._introduction = newRootTree()
+  }
+  set introduction (nodes) {
```
vhf: Using a getter/setter here makes it harder to understand IMO but I might be failing to see the benefits it brings.

`_introduction` is already initialized in the constructor, and later on (for instance l.120) when we assign to `introduction` it's hard to tell whether it enter the setter or not. I would refactor to not use these.

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+    this._introduction = newRootTree(nodes)
+  }
+  get introduction () {
+    return this._introduction
+  }
+  extractConclusions () {
+    this.subTrees.forEach(splittedPart => {
+      const firstHeading = find(splittedPart.children, {
```
vhf: `splitPart`, not `splitted`

packages/mdast-util-split/src/index.js
artragis: this is what I mean. For splitDepth 2 here it what i would like : 

```
{
  introduction: {type: root, children: [/* from doc root to `chapter 1` (not included) */]},
  title: { type: root, children: [/*the heading node*/],
  trees: [
      {type: root, title: {type: 'root', children:[/*chapter 1 heading*/]},
        trees: [/** trees for section1.1=>section 1.2],
        conclusion: {type: root, children:[/* content of section 1.2, but heading discarded*/]},
      {type: root, children: [/* from `chapter 2` to `conclusion` (not included) */]},
  ],
  conclusion: {type: root, children: [/* from `conclusion` (not included) to document end */]},
}
```

packages/mdast-util-split/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+      if (!lastHeading || lastHeading === firstHeading) {
+        return
+      }
+      const rootContent = splittedPart.children.splice(
+        lastIndex,
+        splittedPart.children.length - lastIndex)
+      splittedPart.conclusion = newRootTree(rootContent)
+    })
+  }
+  extractIntroductions () {
```
vhf: In which cases do we have several introductions?

packages/mdast-util-split/src/index.js
vhf: Why does chapter 1 have a conclusion but no introduction? Could you please explain how all of this works in the README?

packages/mdast-util-split/README.md
artragis: if the property set to false, we do not extract the introduction from the tree. In though I had changed like for the conclusion in `if false, the part tree will not be changed`

packages/mdast-util-split/src/index.js
artragis: I think I wanted to use a "simple" get/set at the beginning, failed and instead of reworking the whole code kept the get/set. if you want to remove this, you clearly can, it's an antipattern now.

packages/mdast-util-split/src/index.js
artragis: when all chapters have introductions.

vhf: Do we still extract it when only one chapter has one?

packages/mdast-util-split/src/index.js
vhf: Good, go ahead then

packages/mdast-util-split/src/index.js
artragis: we extract it for every chapter that has one introduction.

artragis:
Thanks to your first review and PR I could have a better view on my work and I was wondering if I was not trying to create a too complex thing.

An idea that showed up was to say : 

just extract one level with two parametters : 

- which depth is considered as "root" depth for titles, default to 1
- do we extract the last paragraph as a conclusion?

the result would just be : 

```
{
    introduction: {type: 'root', children: [everything before first heading]},
    chapters: [all trees],
    conclusion: /* nothing if extractConclusion is false, else the last tree of chapters*/
}
```

thanks to that we can use all the existing utils on subtrees if we want a deeper splitting or a specific treatment.

vhf:
Nice! I agree, the initial code was quite complicated and the purpose unclear, it's a good we now see things more clearly.

> ```
> {
>     introduction: {type: 'root', children: [everything before first heading]},
>     chapters: [all trees],
>     conclusion: /* nothing if extractConclusion is false, else the last tree of chapters*/
> }
> ```

If I understand correctly, when `extractConclusion` is true `conclusion` contains `chapters.pop()`. If that's correct I wouldn't bother, it's not worth it. If anyone wants the last chapter it's easy enough to get it, no need for an extra config option.

Other than that I like the new idea. Mapping the splitter to its result (with depth+1) is good enough when a deeper split is required.

artragis:
> If I understand correctly, when extractConclusion is true conclusion contains chapters.pop(). If that's correct I wouldn't bother, it's not worth it. If anyone wants the last chapter it's easy enough to get it, no need for an extra config option.

in fact I would say `newTree(chapters.pop())` but you are probably right, it's perhaps too much.

> Mapping the splitter to its result (with depth+1) is good enough when a deeper split is required.

perhaps you were rephrasing my message but I do not get it here. What do you mean by "mapping the splitter to its result"?

vhf:
```js
const splitAtDepth = (depth) => (tree) => split(tree, { depth })

const level1 = splitAtDepth(1)(doc)
const level2 = level1.chapters.map(splitAtDepth(2))
```

vhf:
> I'm not sure about the naming of this package. `mdast-util-split` sounds very generic while this package seems to solve a specific thing.

What about `mdast-util-split-by-heading`?

artragis:
About module name : I do not know what is better and having a look at unified documentation I saw you're part of the kind-of-board of unified/remark ecosystem so I think you can put the name you want without negociating it will be the best thing for remark ecosystem.

PR #318, opened 2019-03-03
Ajout d'un plugin filtrant les schemas des URIs

arobase-che:
Coucou \o

Comme promis. Besoin d'une revue. #chezmoiçamarche  
Là seule chose que j'ai pas testé il me semble c'est un réinstallation totale.

À bientôt je pense.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12579,6 +12579,14 @@
         "mdast-util-to-hast": "^3.0.0"
       }
     },
+    "remark-seclink": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/remark-seclink/-/remark-seclink-1.0.0.tgz",
+      "integrity": "sha512-GNd9ttzP4MV7TtuB/+0EBnWOnJCuuvo/8Ji5+Bb8FImF43iIBNlqaSYx2KKykv4RnGSZWGqZU8suHmH7JluCyw==",
+      "requires": {
+        "unist-util-visit": "^1.4.0"
+      }
+    },
     "remark-stringify": {
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/remark-stringify/-/remark-stringify-5.0.0.tgz",
@@ -15302,6 +15310,7 @@
         "remark-parse": "^5.0.0",
         "remark-ping": "file:packages/remark-ping",
         "remark-rehype": "^3.0.0",
+        "remark-seclink": "^1.0.0",
         "remark-sub-super": "file:packages/remark-sub-super",
         "shortid": "^2.2.12",
         "textr": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "remark-kbd": "file:packages/remark-kbd",
     "remark-numbered-footnotes": "file:packages/remark-numbered-footnotes",
     "remark-ping": "file:packages/remark-ping",
+    "remark-seclink": "^1.0.0",
     "remark-sub-super": "file:packages/remark-sub-super",
     "typographic-colon": "file:packages/typographic-colon",
     "typographic-em-dash": "file:packages/typographic-em-dash",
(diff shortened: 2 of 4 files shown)


vhf:
A few comments:

* Nothing gets merged without a clear explanation of why it's needed and how it works.
* Tests need to pass.
* The name of the thing you're trying to add to this project doesn't make enough sense for me to understand the idea.

arobase-che:
Hi \o

The goal of this PR is to protect zmarkdown from XSS URI.
Currently, XSS URI is possible with zmarkdown.

Try that link : 
~~~md
[Cool link try it out](javascript:console.log(alert("Ahaha i'm evil ! 😈  ")))
~~~

I will check the problem of the PR tonight.

vhf:
What is "remark-seclink", how does it prevent the issue?

arobase-che:
It works by whitelisting only some URI schemes.

remark-seclink, is a remark plugin created for this purpose. Thus, i can rename it, or it can be merged into the zestedesavoir repository.

XSS URIs seem to be dangerous security problem.   
I will try to fix the PR a soon as possible. It looks like a dependencies problem. 

Sorry about my english ._.

vhf:
Whitelisting URL schemes is a good approach but the issue you describe has to do with HTML. Some remark plugins generate elements that aren't anchors yet get rendered as HTML anchors. For instance `ping` elements will become HTML anchors and these won't go through your plugin.

HTML anchors are not the only elements for which this type of XSS are possible either, HTML img also seem to be affected: https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#Image_XSS_using_the_JavaScript_directive

For these reasons I don't think this is related to `link` and I don't think this should be a `remark` plugin. If you want to tackle this issue, a rehype plugin would probably be safer and naming it something like `rehype-url-scheme-whitelist` would be much more explicit than `remark-seclink`.

I recommend writing a good README including at least _why (this plugin)_ and _how (this plugin works)_; a good test suite would also be nice.

arobase-che:
Ok.

I have already checked and 
~~~html
<IMG SRC="javascript:alert('XSS');">
~~~
Doesn't seem to work anymore on modern browsers.


They is already a whitelist approach on the media tag : 
~~~md
!(https://www.youtube.com/watch?v=SoLCD3rP9DM)
~~~
A `startBy("http")` is present.

I will follow your advices. Maybe it's important to deal with older browsers who still affected by img's src XSS.

arobase-che:
k

arobase-che:
I will work on a rehype plugin instead of a remark plugin as it's a HTML issue. ;) 

Now that i can see the tests that fails, how should i handle : 
~~~shell
    Expected value to be:
      "<p><a href=\"path/to/image.png\"><img alt=\"altname\" src=\"path/to/img_thumb.png\"></a></p>"
    Received:
      "<p><a href=\"/path/to/image.png\"><img src=\"path/to/img_thumb.png\" alt=\"altname\"></a></p>"
    Difference:
    <p><a href="/path/to/image.png"><img alt="altname" src="path/to/img_thumb.png"></a></p>
~~~

I think, that we should update the tests, shouldn't we ?

A-312:
```js
if (link.url.indexOf(":") !== -1)
```

But this will catch `mypage.html?a=user:1234`, usually ":" should be encoded with : "%20" like this :  `mypage.html?a=user%201234`

An other way : 

```js
function plugin() {
  return transformer;
  function transformer(tree) {
    visit(tree, 'link', link => {
      if (link.url) {
        if (['/', './', '../'].some(link.url[0])) { // Local link
          return;
        }

        try {
          const url = new URL(link.url);

          if (schemes.some(scheme => url.protocol === scheme + ':')) {
            return;
          }
        } catch (e) { }

        link.url = '/' + link.url;
      }
    });
  }
}
```

vhf:
> we should update the tests, shouldn't we ?

I don't think so, they look fine to me. You should only take care of links that have a scheme.

Issue #316, opened 2019-03-01
remark-grid-tables: problem with accented characters


benjaminabel:
Hi, and first, thanks for this handy plugin. I found that the parser do not work when there are accented characters on the text(I'm using it from gatsby). For example:

- this works

```md
+-------+--------+------+
| Table | adings | Here |
+=======+========+======+
| celio | column | ning |
+-------+--------+------+
| rows  | normal | cell |
| rows  | normal | cell |
+-------+--------+------+
```
Bit this with just one more accent **not**:

```md
+-------+--------+------+
| Table | adings | Here |
+=======+========+======+
| célio | column | ning |
+-------+--------+------+
| rows  | normal | cell |
| rows  | normal | cell |
+-------+--------+------+
```

vhf:
Minimal reproduction:

```
+---+
| é |
+---+
```
Interestingly enough it works it doesn't work with `è` either but does with `ö`, `ï`…

vhf:
https://www.npmjs.com/package/remark-grid-tables/v/1.5.0

vhf:
I'm not sure this issue can be 100% resolved because fonts and unicode will never 100% match.

On the one hand grid tables are wysiwyg: the visual representation is supposed to be the truth. On the other hand what unicode says about characters width is not always respected by the fonts.

Take for instance a normal width and two full width:
* ø
* †
* €

They'll often seem to have the same width in monospace fonts though:

```
+---+
| ø |
+---+

+----+
| † |
+----+

+----+
| € |
+----+
```

Not much we can do here IMO.

benjaminabel:
Thanks for the fix, however, it stills don't work with my tables that works with the pandoc [extension-grid_tables](http://pandoc.org/MANUAL.html/grid-tables) extension:

```md
+--------------------+---------------------------------+-------------------------------------+
|      Contenus      |       Capacités attendues       |            Commentaires             |
+====================+=================================+=====================================+
| Événements clés de | Situer dans le temps les        | Ces repères historiques seront      |
| l’histoire de      | principaux événements de        | construits au fur et à mesure de la |
| l’informatique     | l’histoire de l’informatique et | présentation des concepts et        |
|                    | leurs protagonistes.            | techniques.                         |
+--------------------+---------------------------------+-------------------------------------+
```

vhf:
Please provide a minimal reproduction.

benjaminabel:
[...]

PR #319, opened 2019-03-10
remark-grid-tables: better fullwidth support

vhf:
cf. #312 #316 

We had problems with fullwidth support, for instance accented characters such as `é` were being detected as fullwidth, same for Cyrillic script.

@chitoku-k I'd really appreciate your review on this PR since you did the original work to support this and I don't want to break grid tables the scripts / character sets you care about. :)


chitoku-k:
Hi, @vhf.

Looking at the new implementation, I don't see any problems that might break grid tables in full-width support I added. After updating to v1.5.0, I saw a different treatment over `®`, which was previously counted as a halfwidth character. Anyway, as you thankfully described in the issue, it is quite hard to address this sort of ambiguous things. For me, it'd better merge this!

Thanks.

vhf:
Published as https://www.npmjs.com/package/remark-grid-tables/v/1.6.0

PR #320, opened 2019-03-21
zmarkdown: Use rehype-sanitize to prevent dangerous HTML

arobase-che:
Context
=====
Currently, i'm working on fixing URL schemes based XSS like that : 
~~~
[Click me please](javascript:alert("XSS"))
~~~

But @wooorm make me realize, they is a rehype plugin to handle that, [rehype-sanitize](https://github.com/rehypejs/rehype-sanitize). So I simply configure it to work with zmarkdown. 

The aim of [rehype-sanitize](https://github.com/rehypejs/rehype-sanitize) is to deal with "dangerous HTML", URL schemes based XSS seems to be a particular case.

What remains before merging
============

 - **Their is only one test case that fails**. To pass the test, we need to whitelist `file` scheme. I don't think we should do it. The test is about prohibiting the  the `file` scheme. I think that we should update this test, but I will not do it on my own.
 - I don't know if i have to update `packge-lock.json`. I didn't. **Edot** I did, otherwise, travis complains.

💕
ache


vhf:
Nice!

* Updating the failing test to expect `<p></p>` is the right approach, yes
* Please add tests to your PR

vhf:
Thanks @arobase-che !

vhf:
Published: https://www.npmjs.com/package/zmarkdown/v/5.13.0

@Situphen @sandhose feel free to deploy on beta, test and deploy to prod.

arobase-che:
😳 :heart:

Issue #324, opened 2019-05-04
zmardkown/remark-ping: ping-username class is not added


artragis:
in last version of zmarkdown, when we ping a user the ping-username class is not added on span , current result is : 

```html
<a href="/membres/voir/Situphen/" rel="nofollow" class="ping ping-link">@<span>Situphen</span></a>
```
instead of
```html
<a href="/membres/voir/Situphen/" rel="nofollow" class="ping ping-link">@<span class="ping-username">Situphen</span></a>
```

initial markdown whas `@Situphen` @situphen

vhf:
https://www.npmjs.com/package/zmarkdown/v/5.14.0 released

Issue #323, opened 2019-05-04
zmarkdown : as katex dept is not pinned style can easily be broken


artragis:
We updated zmarkdown in production.
But katex team made a recent upgrade to katex 0.10.1. 
We use 0.10.0 style. As katex dependency management is not strict, we now have a parser that generates not compatible css.

vhf:
> As katex dependency management is not strict, we now have a parser that generates not compatible css.
@artragis I guess you meant HTML, not CSS? Could you please point me to a markup difference between `katex@0.10.0` and `0.10.1`?

artragis:
il fact this was... tricky. katex comes with a js depth which is used in rehype-katex. It also comes with a stylesheet used in zds to display equation with the right fonts & borders. It also comes with the fonts.

If a change is made in HTML generation (for example adding a wrapping span or removing a class) this can lead to a not-well-displayed equation. 

We managed to get the "zmarkdown stylesheet" on zds side.

vhf:
Yeah I got that: https://github.com/zestedesavoir/zmarkdown/pull/325#issuecomment-490605274 , that's not what I'm asking for.

We need examples of math things that generate different HTML depending on whether they go through katex 0.10.0 or katex 0.10.1. As soon as we have this we can leverage the markup for tests.

As I said, I don't think the parser generates any CSS; without markup this we won't be able to get tests that would pass for 0.10.0 while failing for 0.10.1.

vhf:
https://www.npmjs.com/package/zmarkdown/v/5.14.0 released

PR #325, opened 2019-05-08
zmarkdown: Resolves conflicts between remark-ping, rehype-katex and rehype-sanitize

TitiAlone:
# Context

On the last stable version of Zeste de Savoir, a bug was reported on ZMarkdown as issue #324 , concerning a bug with remark-ping; some classes in the formatted HTML were missing. The issue is due to a bad configuration of the rehype-sanitize module.

At the same time, a bug with math was reported, and was first believed to be a dependency problem, while it was only rehype-sanitize oversanitizing KaTeX HTML too; this pull request should then also resolve the problem which led to issue #323 , although another step to resolving this issue may be taken later (see below).

# Resolution

To solve these bugs, it was needed to change the rehype-sanitize configuration to:

- allow classes on span elements for both ping and math;
- allow inline styles on span elements for math;
- allow every MathML element to pass through for KaTeX MathML variant.

Also, two tests were written in a new `sanitizer.tests.js` file to check problems with the sanitizer; composed of only two tests (math and ping) today, this file could add more checks for different HTML tags used by various plugins.

# What remains?

- Merging this PR would solve ZMarkdown bugs in production.
- Before doing it, I need someone to check my ping test, which does not fail correctly for some reasons (related with Jest, I think); to test it, please make the ping test fail, and see what happens (tests pass but with Jest error).
- On issue #324 , it will be solved entirely by this pull request.
- On issue #323 , the problem which led to is resolved by the PR, but pinning the dependency somewhere as @artragis suggested could be a wise idea to avoid future conflicts with versions between zds-site and zmarkdown.


packages/zmarkdown/__tests__/sanitizer.tests.js
```diff
@@ -0,0 +1,28 @@
...
+
+const zmarkdown = require('../server')
+
+const renderString = (config = {remarkConfig, rebberConfig}) =>
+  (input) =>
+    zmarkdown(config).renderString(input).then((vfile) =>
+      vfile.toString().trim())
+
+describe('#ping', () => {
+  it('do not oversanitive ping', () => {
```
artragis: oversanitize (v => z)

packages/zmarkdown/__tests__/sanitizer.tests.js
```diff
@@ -0,0 +1,28 @@
...
+    const rendered = renderString()('@Clem')
+    expect(rendered).resolves.toContain('class="ping ping-link"')
```
artragis: should be ok but I don't think class order is specified in HTML

packages/zmarkdown/__tests__/sanitizer.tests.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(rendered).resolves.toContain('class="ping-username"')
+  })
+})
+
+describe('#math', () => {
+  it('do not oversanitize math', () => {
+    const rendered = renderString()('$$1_n$$')
```
artragis: can you add a `\frac{}` command? sub/super and frac were the two most-impacted elements.

packages/zmarkdown/__tests__/sanitizer.tests.js
TitiAlone: What do you mean "class order"? This test only checks for the two classes somewhere in the code, it does not check the order of the two classes, it is not the objective.

Can you run it removing the "l" from class; the test does not fail but it should.

TitiAlone: A lot of tests in the library rely on this ordering, either with `toContain` or `toEqual`.

artragis: ok, so perfect :)

vhf:
Nice, thanks, I'll review this asap!

> On issue #323 , the problem which led to is resolved by the PR, but pinning the dependency somewhere as @artragis suggested could be a wise idea to avoid future conflicts with versions between zds-site and zmarkdown.

That's a weird situation to be in, unfortunately. `zmarkdown` depends on `rehype-katex` which depends on `katex@^0.10.0`. This means that pinning `rehype-katex` in `zmarkdown` won't have any influence on which `katex` version gets pulled as long as it matches `^0.10.0`. If I understand correctly, the issue here is that `katex@0.10.1` got released and someone either deleted the `package-lock.json` on the server or ran `npm update`, otherwise `katex` wouldn't have been updated.

I guess we could add `katex@0.10.0` to `zmarkdown`'s dependencies and pin `rehype-katex` to have `npm` always resolve `katex@^0.10.0` to `katex@0.10.0`? Tell me what you think.

vhf:
The solution I suggested should work, at least locally with a clean lockfile. We need tests for it though. I ran the tests on this PR locally and they are passing with both 0.10.1 and 0.10.0, which means we won't notice if this happens again.

TitiAlone:
> I guess we could add `katex@0.10.0` to `zmarkdown`'s dependencies and pin `rehype-katex` to have `npm` always resolve `katex@^0.10.0` to `katex@0.10.0`? Tell me what you think.

I guess it is the only way to do it if katex is included with a dependency; my point of view is that zds-site should be able somehow to use our dependency, no matter it's version (I think that is what you mentioned on issue #323 , but I don't know if it is possible. The solution of pinning on zds-site would be great because it would allow remark-katex to be updated on the zmarkdown side. But I think the problem was related to old content versions which were broken by updating KaTeX; @artragis will tell us more.

Situphen:
As you said, one solution is to pin KaTeX on both `zmarkdown` and `zds-site` so that they both use the same version.

Another solution is that `zds-site` goes through `zmarkdown`'s `node_modules/` folder to get the files it need. This is what https://github.com/zestedesavoir/zds-site/pull/5321 aims to do. We just need to make sure that zmarkdown is up-to-date before building the front on zds-site, but this is a non-issue in my opinion.

What solution do we choose?

vhf:
Up to you @Situphen .

On zmd side we'll do three things no matter what you decide:
* pin katex to whatever version you want (if 0.10.0 is better for you because of cached generated content, fine by me)
* add tests to make sure katex cannot be upgraded by mistake
* merge this PR

artragis:
What is missing here?

Situphen:
Nothing that I know of. We are just waiting for @vhf to review this PR when he will have the time to do so. :)

vhf:
We are waiting for an answer to this question: https://github.com/zestedesavoir/zmarkdown/issues/323#issuecomment-490610827

Here's why: https://github.com/zestedesavoir/zmarkdown/issues/323#issuecomment-490673187 
> We need examples of math things that generate different HTML depending on whether they go through katex 0.10.0 or katex 0.10.1. As soon as we have this we can leverage the markup for tests.

and then:

> * add tests to make sure katex cannot be upgraded by mistake

cf. https://github.com/zestedesavoir/zmarkdown/pull/325#issuecomment-490974434

artragis:
In fact, having a look at 0.10.0->0.10.1 commits I saw no "html breaking change" BUT even if this is a "bug fix version" there are new features in it so new commands are provided and a "0.10.1-generated html" can be badly displayed with 0.10.0 styles and fonts due to this new commands

image

In zds we clearly prefer 0.10.1 for now.

vhf:
Thanks @TitiAlone! (Awesome PR description!)

TitiAlone:
@vhf : Thanks, I will try to resolve some other issues soon.

What did you do with this:
> Before doing it, I need someone to check my ping test, which does not fail correctly for some reasons (related with Jest, I think); to test it, please make the ping test fail, and see what happens (tests pass but with Jest error).

vhf:
@TitiAlone I couldn't replicate this issue. Putting something random instead of `ping-username` in here for instance:

https://github.com/zestedesavoir/zmarkdown/blob/b62e0da8bc81c428daf876ef9dc0c944a5c91bd7/packages/zmarkdown/__tests__/sanitizer.tests.js#L19

still makes the test fail. It doesn't fail cleanly but it does fail, that's what matters the most.

Here's a way of making it clean:

```js
describe('#ping', () => {
  it('do not oversanitize', async () => {
    const rendered = await renderString()('@Clem')
    expect(rendered).toContain('class="ping ping-link"')
    expect(rendered).toContain('class="ping-username"')
  })
})
```

Issue #326, opened 2019-05-23
zmarkdown: sqrt and overarrows not well interpreted


artragis:
When you use zmd to write math some command generate svg and therefore do not display properly due to sanitizer.

Here are the commands known to generate svg:

```
$$
\overrightarrow v 
\overleftrightarrow v
\overleftarrow v 
\sqrt{x}
\overbrace{x + 1}
\color{x}

\overbrace{x+ 1}
\vec{y}
$$
```

Here is the expected visual result : 
![image](https://user-images.githubusercontent.com/1523653/58248109-8813b680-7d5b-11e9-9a38-2dc28240c8e7.png)

vhf:
Fixed in [zmarkdown@5.14.1](https://www.npmjs.com/package/zmarkdown/v/5.14.1).

PR #327, opened 2019-05-25
zmarkdown: Add more math tests and correct the sqrt problem

TitiAlone:
# Context

After PR #325 , ZMarkdown still generated bad HTML for some math elements; theses have been reported in issue #326 . Eight commands were reported by @artragis to cause problems, and I noted myself that `\\notin` was suspicious also. After a quick investigation, it has been found that theses elements actually needed SVG to be rendered correctly. Unfortunately, SVG is forbidden by the sanitizer by default. This PR tries to solve the problem by changing the sanitizer configuration and adding tests for the problem.

While resolving the issue, I discovered MathML may need [some attributes](https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute) to work correctly; this PR also deals with theses arguments, so that future problems related to this are prevented.

# Resolution

To solve all this, I started by adding tests to see what failed; only four tests were added because all of the nine commands that fail use the same type of tags for rendering. Then, the snapshot were regenerated **after deactivating the sanitizer** and rerun with the sanitizer reactivated to see what was missing.

Then, the KaTeX sanitizer configuration, which would become huge, was moved to an external file for reasons of clarity. The sanitizer configuration now depends on three elements: GH markdown, local KaTeX configuration and default configuration for ZMd.

Finally, the missing HTML elements and attributes were added to the new configuration file, along with the MathML attributes listed above, and test were ran again, but passing this time!

# What remains

- Merging this PR would solve ZMarkdown bugs in production.
- The math issue is now fixed, and we should see no more KaTeX problems... until the next version.
- Issue #326 would be solved.


vhf:
That's great, thanks!

> The math issue is now fixed, and we should see no more KaTeX problems... until the next version.

Hopefully that won't happen by accident, I pinned katex@0.10.1 here: a56426c54291462ebeeef7b848c016fd6c719a30

artragis:
Thanks for pinning. 0.10.2 is now out and include many changes. If we want it to go live on zmd we must perform beta test before.

vhf:
Published [zmarkdown@5.14.1](https://www.npmjs.com/package/zmarkdown/v/5.14.1).

> If we want it to go live on zmd we must perform beta test before.

👍

PR #333, opened 2019-06-16
zmarkdown: Add statistics facility

artragis:
# goal of this PR

On ZDS we have an estimated time of reading for all contents.

For now this indication is not reliable at all because it uses a bare `len(text)` to evaluate the size of content.
This is problematic because it includes mardown-syntax and URL related chars.

In this PR I evaluate the true number of chars by visiting text nodes.

### DIFF
--- a/packages/zmarkdown/README.md
+++ b/packages/zmarkdown/README.md
@@ -56,6 +56,10 @@ Only `metadata` is described in the **Response** sections below.
 
   Only parse inline Markdown elements (such as links and emphasis, unlike lists and fenced code blocks)
 
+* `opts.stats`, bool, default: `false`
+
+  Will compute and return statistics about markdown text
+ 
 ### Response
 
 * `metadata.disableToc`, bool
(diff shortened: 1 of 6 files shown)


packages/zmarkdown/README.md
```diff
@@ -73,6 +77,12 @@ Only `metadata` is described in the **Response** sections below.
 
   A list of unique languages used in GitHub Flavoured Markdown fences with a flag.
 
+* `metadata.stats` , object
+
+  stats about the parsed text:
+   - `signs`: number of chars, spaces uncluded
```
vhf:
```suggestion
   - `signs`: number of chars, spaces included
```

artragis:
You've got a point

I will change it asap.

vhf:
> I think `signs += node.value.length + 1` wouldn't help because `a *b* c` would be 8 instead of 5, right?

As expect, 45307bc gives `8` chars for `a *b* c`. I think that's too far off.

> What about adding something like
> 
> ```js
>       visit(tree, 'paragraph', (node) => {
>         signs += 1
>       })
> ```
> 
> before or after visiting `text`?

Your thoughts on this?

artragis:
I just messed up.

artragis:
should be good now

packages/zmarkdown/README.md
vhf: @artragis

vhf:
🚀 https://www.npmjs.com/package/zmarkdown/v/5.15.0

Issue #314, opened 2019-02-13
Add Soundcloud


A-312:
Add soundcloud like youtube video : 

    !(https://soundcloud.com/raz3c21/airplane-hayley-ft-b-o-b

HTML : 

     <iframe width="100%" height="300" scrolling="no" frameborder="no" allow="autoplay" src="https://w.soundcloud.com/player/?url=https%3A//api.soundcloud.com/tracks/6038702&color=%23ff5500&auto_play=false&hide_related=false&show_comments=true&show_user=true&show_reposts=false&show_teaser=true&visual=true"></iframe>

artragis:
as @TitiAlone merged oEmbed we can close, can't we?

vhf:
I'll close with the package version once released :)

vhf:
remark-iframes@4.0.0 / zmarkdown@6.0.0

PR #334, opened 2019-06-19
[remark-iframes] Add oEmbed functionnality

TitiAlone:
Hello there,

this is a draft PR about adding automatic iframe description using the [oEmbed APIs](https://oembed.com/) of the various websites we need.

# Why this PR?

I wanted to work on Soundcloud integration to ZMarkdown as requested on #314 and I noticed Soundcloud integration is not possible to do with the current remark-iframes configuration, as too many changes are involved for formatting the URL, and we would need full RegEx control on the configuration.

As I would have found this configuration rather dirty, I prefered to use oEmbed APIs of the various websites we use that implement it. The global principle is to make a request to the website which will return the integration URL we need.

## What's good?

- really easy to add new content providers, even in cases when the configuration would be non-trivial;
- the system stay decentralized, as every content provider has it's own API, we do not rely on a third-party service (other than the one we include content for).

## What's not so good?

- requests need to be done synchronously, so there may be an small delay between request and response for the ZMarkdown server;
- if the service is down, not integration will be provided, but do we really need to handle this case?

# Do not merge before

This is only a demonstration in order to get some feedback about the principle; please do not merge before the following issues aren't resolved:

- tests are added for unreachable content (ex. Youtube: BpJKvrjLUp0);
- an error is triggered when request cannot succeed;
- Soundcloud integration is provided;
- requests are made asynchronously;
- a correct fallback is in place;
- code is reviewed, obviously.


packages/remark-iframes/src/index.js
```diff
@@ -40,23 +41,56 @@ module.exports = function plugin (opts) {
...
-        hProperties: {
-          src: finalUrl,
-          width: provider.width,
-          height: provider.height,
-          allowfullscreen: true,
-          frameborder: '0',
+    let finalUrl, thumbnail, fallback
+
+    if (provider.oembed) {
+      const reqUrl = provider.oembed + '?format=json&url=' + encodeURIComponent(url)
```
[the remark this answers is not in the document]
TitiAlone: Sorry about me forgetting to enable the linter.
I've got no idea what this warning is about...

vhf:
(For some reason github didn't let me answer here: https://github.com/zestedesavoir/zmarkdown/pull/334#discussion_r295395685 )

```suggestion
      const reqUrl = `${provider.oembed}?format=json&url=${encodeURIComponent(url)}`
```
> Unexpected string concatenation prefer-template

The linter noticed you were concatenating strings (`'a' + b + 'c'`), it expects the codebase not to have string concatenation. It prefers template strings. :)

TitiAlone:
Now tests are passing, @vhf could I get your opinion about this PR?

vhf:
It's looking good, oembed is obviously a better solution! 👍 

> ## What's not so good?
> * requests need to be done synchronously, so there may be an small delay between request and response for the ZMarkdown server;

Does it though? Visitors can be asynchronous, see `remark-images-download` for example.

> * if the service is down, not integration will be provided, but do we really need to handle this case?

We do :)  
A simple timeout would do. I expect oembed APIs to be quite responsive, and we don't want to block markdown rendering for too long anyway. Something like 3000ms per request to an oembed API endpoint?

TitiAlone:
> We do :)
> A simple timeout would do. I expect oembed APIs to be quite responsive, and we don't want to block markdown rendering for too long anyway. Something like 3000ms per request to an oembed API endpoint?

But what to do in case the timeout has elapsed? Fallback to the old solution? It seems rather dirty...

vhf:
We don't have to, we could simply render a link instead of the iframe. What's for sure is that we cannot block the whole zmarkdown server indefinitely. Making this asynchronous won't block the server for other clients but still, it needs to render in a timely manner.

TitiAlone:
I've got a problem with UnhandledPromiseRejectionWarning; it tells me I need to handle some kind of Promise on the `eat`, and I don't know why there should be a Promise handling here; when this problem will be resolved, asynchronous requests will likely be working.

Any clue @vhf ?

vhf:
Yep, that's not the right approach, way too complicated. I'll PR your branch with a fix.

Basically:
* a tokenizer cannot be async
* using `node-fetch` makes much more sense than manually building a request
* there's a confusion with async and promises here, a weird mix ;)

What I'll work on:
* the tokenizer always falls back and puts the necessary data for the fetch call in data attributes
* an async visitor goes through the relevant nodes and replaces with the fetch result

vhf:
Here it is: https://github.com/TitiAlone/zmarkdown/pull/1 

Merging it will automatically add my commit to your PR. You can then remove the "Draft" status from this PR (no idea how, first time I see this thing) and we'll be good to go

TitiAlone:
This is great, thanks a lot.

I will read your code and comments tomorrow, and merge it right after. Anyhow, do you have a link to some unified documentation (about how processing is done, especially)? I couldn't find anything apart from their (very light) guide.

vhf:
> Anyhow, do you have a link to some unified documentation (about how processing is done, especially)? I couldn't find anything apart from their (very light) guide.

Unfortunately I don't. The best I can provide is:
* phases/process: https://github.com/unifiedjs/unified#description
* transformer: https://github.com/unifiedjs/unified#function-transformernode-file-next

Consider a plugin such as [this one](https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/src/index.js#L5) and run [like this](https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/__tests__/index.js#L11-L16):

```js
const render = async (text, config) => unified()
  .use(reParse) // parsing
  .use(plugin, config)
  .use(remark2rehype) // mdast to hast
  .use(stringify) // hast to html
  .process(text)
```

1. It defines a new block tokenizer: https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/src/index.js#L15-L90 and adds it to `remark-parse` existing tokenizers: https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/src/index.js#L95-L98 . This will be used in `Parser` phase of the process.
2. It also detects whether the plugin is run during a `Compiler` phase of the process, in this case it tells the compiler how to compile to markdown: https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/src/index.js#L100-L105
3. It returns a `Transformer`: https://github.com/TitiAlone/zmarkdown/blob/8d00ffcb9b6de567f3b95852a1a39ed5d2b2df84/packages/remark-iframes/src/index.js#L107-L158 , which will be called after the whole `Parser` phase is done. Since this is a remark plugin, the transformer runs on mdast: the markdown AST representation of the input markdown document.

TitiAlone:
Everything is ready to be merged!

What I did on my last commit is to add Soundcloud provider, which is what this PR was really about; a test was also added to make sure everything would work perfectly.

packages/remark-iframes/src/index.js
```diff
@@ -127,3 +210,18 @@ function computeThumbnail (provider, url) {
   }
   return thumbnailURL
 }
+
+async function fetchEmbed (url) {
+  return fetch(url, {timeout: 1500})
+    .then((res) => res.json())
+    .then((oembedRes) => {
+      const oembedUrl = oembedRes.html.match(/src="([A-Za-z0-9_/?&%=:.]+)"/)[1]
```
vhf: What about:
```suggestion
      const oembedUrl = oembedRes.html.match(/src="([^"]+)"/)[1]
```

TitiAlone: We can do that if you want, but there's no need to exclude ", we could just:

```javascript
const oembedUrl = oembedRes.html.match(/src="(.+?)"/)[1]
```

vhf: Nice, let's do this. I cannot apply the suggestion because it's your branch ;)

TitiAlone:
Now we should be ready to merge @vhf .

PR #336, opened 2019-06-25
rebber-plugins: change column rendering

### LINKED ISSUE #335: rebber-plugins: No centering inside table
LaTeX does not support allignement inside table. Thus,

```
+-------+----------+------+
| Table Headings   | Here |
+=======+==========+======+
| rows  | normal   | cell |
+-------+----------+------+
| multi | ->cells can be  |
| line  | *formatted*     |
|       | **paragraphs**  |
| cells | <-              |
| too   |                 |
+-------+-----------------+
```

Is [actually translated](https://zestedesavoir.github.io/zmarkdown/) into

```latex
\begin{longtabu}{|p{\linewidth / 3}|p{\linewidth / 3}|p{\linewidth / 3}|} \hline
\rowfont[c]{\bfseries}
\multicolumn{2}{|c|}{Table Headings} & Here \\ \hline
\rowfont[c]{\bfseries}
rows & normal & cell \\ \hline
\rowfont[l]{}
multi \par line \par  \par cells \par too & \multicolumn{2}{|c|}{\begin{center} \par cells can be \par \textit{formatted} \par \textbf{paragraphs} \par  \par  \par \end{center}} \\ \hline
\end{longtabu}
```

But `\begin{center}` cannot be inside a table (also, the `{|c|}` of `\multicolumn` already centers the content of the cell).

artragis:
fix #335 

Remaining : 

- [x] generate dist file, @vhf I've got errors when I run npm run prepare
- [x] @karnaj @pierre-24 are the columns rendered as wanted?
- [x] @karnaj @pierre-24 is this ok for multi row?


packages/rebber-plugins/src/type/gridTable.js
```diff
@@ -112,7 +114,7 @@ class GridTableStringifier {
       this.nbOfColumns = this.colIndex
     }
     this.colIndex = 0
-    return rowText
+    return rowText.replace(/\\number-of-column/, this.nbOfColumns)
```
vhf: This should only replace the first occurrence. Is this really what you meant? If so please remove the regex.

artragis: At the begining it was not what I needed, I just forgot the `g` option.

But in fact, this can only be found on the first row so I don't know if the regexp is really the right thing to do

vhf:
I'll take a look when we're sure that's what we want. 👍

artragis:
So, I had some time to think about it : here is the problem : 

We want to change the way we provide table header to make latex be able to use alignment AND dynamic sizing.

For that we have to use the `p{\dimexpr(\linewidth) * {colspan} / {number of column}}` command.
However the "number of column" is dynamicaly computed so on the first column we cannot use "number of column" when we format the string. 

That's why I used the "placeholder" technique : for each line, once we are sure number of column is computed, we replace all reference to `\number-of-column`. In each line we can have as many `\number-of-column` as we have columns.

If you think the technique is good, we just have to add the `g` parameter in regexp. Otherwise feel free to fix my method.

artragis:
So it's ok for me, the main issue is that I cannot run npm prepare (babel fails at home). 

```
fdambrine@DLIL0005:/mnt/c/Users/fdambrine/perso/zmarkdown/packages/rebber-plugins$ npm install

> rebber-plugins@1.5.7 prepare /mnt/c/Users/fdambrine/perso/zmarkdown/packages/rebber-plugins
> del-cli dist && cross-env BABEL_ENV=production babel src --out-dir dist

Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel. (While processing preset: "/mnt/c/Users/fdambrine/perso/zmarkdown/node_modules/@babel/preset-env/lib/index.js")
    at throwVersionError (/mnt/c/Users/fdambrine/perso/zmarkdown/node_modules/@babel/helper-plugin-utils/lib/index.js:65:11)
```

artragis:
thanks @TitiAlone @vhf once you're Ok it can be merged.

vhf:
@artragis Could you please rebase to get rid of all unrelated commits?

artragis:
should be good now.

PR #337, opened 2019-07-01
Fix regex to correctly match brackets

itsknob:
Regex had an unescaped closing bracket causing issues when searching
for special characters that need to be escaped by the replace function.

Was only matching `-/{}()*+?.\^$|]` as a whole because of closing bracket, now matches single character from list, replacing globally.


itsknob:
I screwed this up pretty good. I'll try again later.

PR #338, opened 2019-07-02
Update regex to correctly match brackets

itsknob:
Trying again.
Remove unnecessary escapes. Add necessary escape to correctly 
match brackets. Pattern was incorrectly matching `-` or `[`
followed by the literal string `/{}()*+?.\^$|]` only correctly
replacing the `-` or the `[` with a escape `\`.

### DIFF
--- a/packages/remark-custom-blocks/src/index.js
+++ b/packages/remark-custom-blocks/src/index.js
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(/[-[]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
+  return str.replace(/[-[\]{}()/*+?.\\^$|]/g, '\\$&')
 }
 
 const C_NEWLINE = '\n'
--- a/packages/remark-custom-blocks/src/index.js
+++ b/packages/remark-custom-blocks/src/index.js
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'))
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'), "\\$&")
 }
 
 const C_NEWLINE = '\n'
--- a/packages/remark-custom-blocks/src/index.js
+++ b/packages/remark-custom-blocks/src/index.js
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'), "\\$&")
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'), '\\$&')
 }
 
 const C_NEWLINE = '\n'
--- a/packages/remark-custom-blocks/src/index.js
+++ b/packages/remark-custom-blocks/src/index.js
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(/[-[\]{}()/*+?.\\^$|]/g, '\\$&')
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'))
 }
 
 const C_NEWLINE = '\n'


itsknob:
This solution does not escape the forward slash. Will update soon.

vhf:
Hey @itsknob , thanks! Could you please explain what this PR solves? Did you encounter issues with `remark-custom-blocks`?

It would be good to have tests for this as well, I'm happy to take care of it if you don't want to, as long as I can reproduce the issues you had.

packages/remark-custom-blocks/src/index.js
```diff
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(/[-[]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'))
```
vhf: This would replace any match with `undefined`:

```js
> escapeRegExp('[AA-AA|AA]AA')
'undefinedAAundefinedAAundefinedAAundefinedAA'
```

packages/remark-custom-blocks/src/index.js
```diff
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(/[-[]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'), "\\$&")
```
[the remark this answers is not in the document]
itsknob: > This would replace any match with `undefined`:
> 
> ```js
> > escapeRegExp('[AA-AA|AA]AA')
> 'undefinedAAundefinedAAundefinedAAundefinedAA'
> ```

I forgot to add the `\\&$` for the replacement string. I've updated the commit.

image

packages/remark-custom-blocks/src/index.js
itsknob: I've updated the replace string in a commit, I had forgotten it.

itsknob:
> Hey @itsknob , thanks! Could you please explain what this PR solves? Did you encounter issues with `remark-custom-blocks`?
> 
> It would be good to have tests for this as well, I'm happy to take care of it if you don't want to, as long as I can reproduce the issues you had.

Honestly, I was digging into [this issue](https://github.com/gatsbyjs/gatsby/issues/15057) for the Gatsby repo, and I may have dug a little to deep when I noticed the regex. I thought it may have something to do with when the tag name is not matched correctly. I pulled the regex out and realized it was not matching correctly. [Shown here](https://regex101.com/r/hMtuZV/1)

packages/remark-custom-blocks/src/index.js
```diff
@@ -1,7 +1,7 @@
 const spaceSeparated = require('space-separated-tokens')
 
 function escapeRegExp (str) {
-  return str.replace(/[-[]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
+  return str.replace(new RegExp(`[-[\\]{}()*+?.\\\\^$|/]`, 'g'), '\\$&')
```
vhf: `()` is an invalid capturing group, then `*+?` are three consecutive quantifiers that aren't quantifying anything, they all need to be escaped. `.` also needs to be escaped. Then `|` also needs to be escaped, otherwise it's a logical OR.
`/` doesn't *need* to be escaped but it wouldn't hurt to do so.

vhf: To me the correct fix here would be:
```diff
- [-[]\/\{\}\(\)\*\+\?\.\\\^\$\|]
+ [-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]
```

vhf: I used https://regexr.com to check your regex and I think this website got it wrong. Now I'm not sure what is correct and what isn't. I wrote tests and it seems that your regex works properly.

vhf:
Alright, I'm not sure fixing this here will fix the issue you were investigating but it's still worth fixing.

Could you please add this test case to the end of `packages/remark-custom-blocks/__tests__/index.js`?

It tests this specific bug since this crashes on `master` but passes on your PR:
```js
test('weirdly named blocks', () => {
  // a block name would usually be something like "info" or "spoiler"
  // here we test that using special characters, specifically regex
  // special characters, still works as block names since they are
  // correctly escaped
  const blockNames = [
    '[a]',
    '(b|)',
    '[a})',
    '[a-Z]',
    '.*',
    '??}',
    '^$',
    '|||',
    ']',
    '[]]',
  ]
  const config = blockNames.reduce((acc, blockName, idx) => {
    acc[blockName] = {
      classes: `class${idx}`,
    }
    return acc
  }, {})
  const render = (text, allowTitle) => unified()
    .use(reParse)
    .use(remark2rehype)
    .use(plugin, config, allowTitle)
    .use(stringify)
    .processSync(text)

  const makeFixture = (blockName) => dedent`
    [[${blockName}]]
    |     [[${blockName}]]
    |     | this
  `

  blockNames.forEach((blockName, idx) => {
    const {contents} = render(makeFixture(blockName))
    expect(contents).toContain(`class="custom-block class${idx}"`)
    expect(contents).toContain(`<pre><code>[[${blockName}]]`)
  })
})
```

packages/remark-custom-blocks/src/index.js
itsknob: I utilized both https://regexr.com and https://regex101.com. I believe I had better results on the later.
I was having trouble with the ESLint rule for unnecessary escapes when escaping all of the characters, and trouble with the regex itself when only escaping the necessary ones utilizing the regex literal, which is why I moved to the RegExp class.

itsknob:
I'll add the tests today, if I can get to it.

vhf:
No worries I'll take care of the tests, thanks for your contribution!

vhf:
Published: remark-custom-blocks@2.4.0

PR #342, opened 2019-07-20
Updating remark, unified etc

vhf:
This branch is tracking @arobase-che 's progress on upgrading remark, unified, etc


artragis:
Can I help here for somthing?

arobase-che:
Hi, I'm still working on it.

I got a weird case.

Since : https://github.com/remarkjs/remark/pull/424

Spaces in inline code are preserved. So a test case doesn't pass.

The test is : 
~~~md
New line: `\
`
The old result was « nothing ». Since the inline code became empty it was deleted.
Now, the backslash is escaped and the new line remains inside the inline code.

I don't understand the new behavior. I think it's not a useful test case and we should just ignore it. I have plenty of test case to study.

Can I just skip it and update the snapshot ? I will leave a issue on remark. GFM and CommonMark replace the new line by a space. Nothing interesting since the chose of remark seems legit.

vhf:
I'm not sure I understand what the issue is, to me remark seems to default to the correct behavior. Apparently you are seeing something different that is not consistent with [`gfm: true`](https://github.com/zestedesavoir/zmarkdown/blob/b05f25a9f9f7be3c2987fb3c07f48cc1b43ca2aa/packages/zmarkdown/config/remark.js#L51)?

Here's how the case you present should be rendered IMO:

~~~md
New line: `\
`
~~~
=>
```html
<p>New line: <code>\ </code></p>
```

Basic example showcasing this behavior: https://github.com/vhf/remark-playground

arobase-che:
> I'm not sure I understand what the issue is, to me remark seems to default to the correct behavior. Apparently you are seeing something different that is not consistent with [`gfm: true`](https://github.com/zestedesavoir/zmarkdown/blob/b05f25a9f9f7be3c2987fb3c07f48cc1b43ca2aa/packages/zmarkdown/config/remark.js#L51)?

Yeap ! Here an example : https://github.com/arobase-che/remark-playground/tree/parsed

I don't know why your version works but mine is problematic with rebber.
https://github.com/arobase-che/remark-playground/tree/rebber

> Here's how the case you present should be rendered IMO:
> 
> ```lisp
> New line: `\
> `
> ```
> 
> =>
> 
> ```
> <p>New line: <code>\ </code></p>
> ```
> 
> Oct-03-2019 22-17-51
> 
> Basic example showing this behavior: https://github.com/vhf/remark-playground . Using `.use(markdown, {gfm: true})` instead of https://github.com/vhf/remark-playground/blob/6f200e3a195f03453c5ffbf375619378430b0ba1/index.js#L9 doesn't change anything.

I totaly agree with you ^^

arobase-che:
Btw, I'm suppose to edit the fixtures by hand ?

vhf:
> I don't know why your version works but mine is problematic with rebber.
> https://github.com/arobase-che/remark-playground/tree/rebber

Here's my example again:

```js
const inspect = require('util').inspect
const vfile = require('to-vfile')
const report = require('vfile-reporter')
const unified = require('unified')
const markdown = require('remark-parse')
const remark2rehype = require('remark-rehype')
const html = require('rehype-stringify')

unified()
  .use(markdown, {gfm: true})
  .use(() => (vfile) => { console.warn(inspect(vfile, false, 9999, true)) })
  .use(remark2rehype)
  .use(() => (vfile) => { console.warn(inspect(vfile, false, 9999, true)) })
  .use(html)
  .process(vfile.readSync('example.md'), function(err, file) {
    console.error(report(err || file))
    console.log(String(file))
  })
```

We see that `remark-parse` keeps \n:
```
        {
          type: 'inlineCode',
          value: '\\\n',
```
and then `remark-rehype` removes it:
```
        {
          type: 'element',
          tagName: 'code',
          properties: {},
          children: [ { type: 'text', value: '\\ ' } ],
```

* https://github.com/remarkjs/remark-rehype/blob/47e547d20a4e79c64b211c92a622042a691237ed/index.js#L3
* https://github.com/syntax-tree/mdast-util-to-hast/blob/a6135c5f478027298ae4bbd26e8f1279a3fead69/lib/handlers/inline-code.js#L5

I'll check all the places where `collapse-white-space` is used in `mdast-util-to-hast` and apply the same to the corresponding rebber/rebber-plugins code. :)



> Btw, I'm suppose to edit the fixtures by hand ?

That's probably the safest thing to do, yes. Are there a lot of HTML changes? Will the HTML changes cause issues on zds?

arobase-che:
Ok :)

> > Btw, I'm suppose to edit the fixtures by hand ?
>
> That's probably the safest thing to do, yes. Are there a lot of HTML changes? Will the HTML changes cause issues on zds?

Enough to be upsetting  :( 
Usually, it's a correction of remark. Half of them are about footnotes (paragraphs and numbers) or consecutives quotes that won't merge now.

Example: 

~~~md
> Here is one

> Here is another and not the same
~~~

vhf:
> > Are there a lot of HTML changes? Will the HTML changes cause issues on zds?
> 
> Enough to be upsetting :(

Then please make sure people agree with these changes before we land this PR :)

artragis:
FOr quotes it clearly is what we want.

Issue #126, opened 2017-07-26
IDL descriptions for mdast extensions


wooorm:
I suggest documenting in each project that adds new mdast nodes how they work, in some way similar to the [mdast spec](https://github.com/syntax-tree/mdast#ast).

vhf:
Progress:

* [ ] [`remark-abbr`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-abbr)
* [ ] [`remark-align`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-align)
* [ ] [`remark-captions`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-captions)
* [ ] [`remark-comments`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-comments)
* [ ] [`remark-custom-blocks`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-custom-blocks)
* [ ] [`remark-disable-tokenizers`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-disable-tokenizers)
* [ ] [`remark-emoticons`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-emoticons)
* [ ] [`remark-escape-escaped`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-escape-escaped)
* [ ] [`remark-fix-guillemets`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-fix-guillemets)
* [ ] [`remark-grid-tables`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-grid-tables)
* [x] [`remark-heading-shift`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-heading-shift)
* [x] [`remark-heading-trailing-spaces`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-heading-trailing-spaces)
* [ ] [`remark-iframes`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-iframes)
* [x] [`remark-kbd`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-kbd)
* [ ] [`remark-numbered-footnotes`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-numbered-footnotes)
[...]

PR #344, opened 2019-07-28
[remark-emoticons] Add a new locator for faster parsing and [zmarkdown] new emoticons

TitiAlone:
Hello!

# Context

New emoticons have been available for Zeste de Savoir for a long time, so I guess now is time to integrate these into ZMarkdown. This PR is the second one that aims to do so, but the first one with all tests running (only three snapshots updated). I'm reposting here the work done on @A-312 's fork, because he asked me to do so to resolve conflicts between the base master and his branch. This PR also includes a new `locator` as mentioned below.

# Work done

I also worked on a new locator for remark-emoticons, searching for chars instead of using Regex; it is way more clear, and should also be faster, to be able to search for all our new emojis. Tests are now up and running, and so we should be able to merge it after a quick review. By the way, I cleaned the AST IDL description to make it unambiguous, and it can be referenced on the related issue #126 .

**The tokenizer was left as is by now**, maybe we could rewrite it later, but for now, we're not parsing that much emoticons. By the way, I cleaned the AST IDL description to make it unambiguous, and it should be referenced on the reference issue #126 .

The previous work from @A-312 is left as it was made, because he knows better than me how the workflow is don on zds-site. For once, nothing left to do, except for a review (poke @vhf ), that I hope will not find too much mistakes .


TitiAlone:
No idea why, tests do not pass on Node 8 for a seemingly unrelated problem...

TitiAlone:
I think we should wait for the new ZdS version; @artragis said he would get these deployed with the upcoming ZdS version, so maybe in some months (we don't really know, actually, but could be long). Releasing it now would be blocking for other PRs (I mean that they couldn't be deployed on ZdS); I don't think it would be wise to merge this now, but at least it's here for whenever it become necessary.

vhf:
@TitiAlone any progress on this front?

A-312:
@artragis can answer

artragis:
zds-site hasn't the same release rythm and cycle than zmarkdown. The main issue here is : if we merge this PR now, we won't be able to take advantage of fixes and functionalities because we can't deploy it on production.

One solution could be a configuration file/env variable that will allow us to tell the service the current emoticon are png ones and later we will use the svg.

Without that we can't deploy any version of zmarkdown defining non-existing emots.

artragis:
@vhf you can merge this PR when you feel zmd is ready to it. We are ready to use what's needed on zds-site side.

PR #347, opened 2019-08-06
Add IDL description for packages where it was missing

TitiAlone:
Solves issue #126 ; IDL description was added for

- remark-align;
- remark-comments;
- remark-custom-blocks;
- remark-grid-tables.

It has already been done by some previous PRs for:

- remark-abbr and
- remark-emoticons.

I also added descriptions in the main README for packages where it was missing.

Two questions for me:

- is there a way to use something like "div OR details" instead of "string" for rehype's `hName`?
- maybe we could do something better for my `whatnotCustomBlock`?

Poke @vhf for a review.

### DIFF
--- a/packages/remark-custom-blocks/README.md
+++ b/packages/remark-custom-blocks/README.md
@@ -23,7 +23,7 @@ By default, the plugin will produce the following two nodes, where `whatnot` is 
 interface whatnotCustomBlock <: Parent {
   type: "whatnotCustomBlock";
   data: {
-    hName: string;
+    hName: "div" or "details";
     hProperties: {
       className: [string];
     }
@@ -35,7 +35,7 @@ interface whatnotCustomBlock <: Parent {
 interface whatnotCustomBlockBody <: Parent {
   type: "whatnotCustomBlockBody";
   data: {
-    hName: string;
+    hName: "div";
     hProperties: {
       className: [string];
     }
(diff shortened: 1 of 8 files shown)


vhf:
> * is there a way to use something like "div OR details" instead of "string" for rehype's `hName`?

It's a WebIDL-like syntax, not directly WebIDL, there's no formal spec to my knowledge. You could take inspiration from WebIDL's union types: https://heycam.github.io/webidl/#idl-union and use `or`

> * maybe we could do something better for my `whatnotCustomBlock`?

I don't have a better idea, feel free to make it better though :)

TitiAlone:
I added the "or" IDL-like syntax.

> > * maybe we could do something better for my `whatnotCustomBlock`?
> 
> I don't have a better idea, feel free to make it better though :)

The question was more about have an IDL-improvement than doing the funniest name possible :P ; but it seems there is no way to make it *better*.

packages/remark-custom-blocks/README.md
```diff
@@ -15,6 +15,48 @@ Each custom block can specify CSS classes and whether users are allowed or requi
...
+  data: {
+    hName: "div";
+    hProperties: {
+      className: [string];
+    }
+  }
+}
+```
+
+If your block has an heading, the following node will also be produced:
```
vhf:
```suggestion
If your block has a heading, the following node will also be produced:
```

vhf:
Thanks a lot!

Issue #349, opened 2019-08-09
rebber-plugins: protect chars in custom blocks with titles


artragis:
# configuration

## remark
```javascript
information: {
      classes: 'custom-block-information',
      title: 'optional',
    },
}
```
## rebber

```
informationCustomBlock: require('rebber-plugins/dist/type/customBlocks'),
```

# markdown

```markdown
[[information| Github, the microsoft code forge]]
| this is github
```

# rendered latex

```tex
\begin{Intormation}[Github, the microsoft code forge]
this is github
\end{Information}
```

# expected
```tex
\begin{Intormation}[Github\protect{,} the microsoft code forge]
this is github
\end{Information}
```

pierre-24:
Yep :)

artragis:
is comma the only char to protect?

pierre-24:
same ass #276 : also square brackets :)

artragis:
according to karnaj we can protect by wrapping the caption between `{{` and `}}`

TitiAlone:
@pierre-24 : Is the resolution done at #279 valid for this case (adding the caption under the block)?

If not, could you confirm `{{` and `}}` would work?

pierre-24:
Because it is not a caption that you put under something (like a table, an image or a code), so the resolution of #279 will not work. But `{{,}}` works, just tested with:

```
\begin{Information}[Github{{,}} the microsoft code forge {{[}}or is it?{{]}}]
this is github
\end{Information}
```

TitiAlone:
@pierre-24 : If I understood well what Karnaj suggested, he thinks that:

```
\begin{Information}[{{Github, the microsoft code forge [or is it?]}}]
this is github
\end{Information}
```

would work. Can you confirm this, because it would be much easier to implement?

pierre-24:
It does work! I feel a bit stupid not to have thought about that ^^

TitiAlone:
I solved the problem quickly, but I'm wondering about:

```
\begin{Information}[{{Github, the microsoft \textbf{code forge} [or is it?]}}]
this is github
\end{Information}
```

Would the command in the title be executed properly (`textbf`)?

pierre-24:
![Screenshot from 2019-08-10 21-19-41](https://user-images.githubusercontent.com/7889675/62826001-c67b7d80-bbb4-11e9-9522-bdfd31b1052c.png)

No problem :)

TitiAlone:
[...]

PR #351, opened 2019-08-10
rebber-plugins: Protect custom block titles

TitiAlone:
Simply adds `{{` and `}}` on custom block titles for LaTeX rendering in order to prevent problems during the PDF generation. See issue #349 for details of the solution. Tests already handle this, so snapshots were only updated.

~~Please wait for confirmation on the issue before merging.~~
It works, no problems whatsoever.


vhf:
Thanks!

PR #353, opened 2019-08-13
Solve the new \vec problem

TitiAlone:
# Context

Once more, a bug was found on a LaTeX command; after PR #327 , ZMarkdown was still generating bad HTML for the `\vec` command; it have been reported on the forum this morning.

# Resolution

I always proceed as follow for theses issues:

1. write a simple unit test;
2. comment the rehype-sanitize line on the `common.js` file;
3. generate the snapshot;
4. uncomment the line;
5. run the test and compare diffs;
6. add the missing element or attribute;
7. repeat from 5.

Here, only one attribute was missing: `style` on `svg`; the command generating the issue now has a test, and the sanitizer has been reconfigured one more.

Nothing left to do, except for merging (and publishing).


vhf:
Great, thanks! (The process listed in *Resolution* is 👌)

vhf:
This has been released as:

- zmarkdown@6.1.0

Issue #354, opened 2019-08-13
zmarkdown & rebber & rebber-plugins: images in heading, table & gridTables should be inline images


artragis:
# Given the markdown

```
# ![image](https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png)

+-----------+------------------------------------------------------------------------------------------------+
| cell      |![img](https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png)  |
+-----------+------------------------------------------------------------------------------------------------+


c1 | c2
---|--
c3 | ![image](https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png)

```

# I get

```
\levelOneTitle{\image{https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png}}


\begin{longtabu}{|p{\dimexpr(\linewidth) / 2}|p{\dimexpr(\linewidth) / 2}|} \hline
\rowfont[c]{\bfseries}
cell & \image{https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png}[img] \\ \hline
\end{longtabu}


\begin{longtabu} spread 0pt {|X[-1]|X[-1]|} \hline
\rowfont[c]{\bfseries}
c1 & c2 \\ \hline
\rowfont[l]{}
c3 & \image{https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png} \\ \hline
\end{longtabu}


```

# instead of

```
\levelOneTitle{\inlineImage{https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png}}


\begin{longtabu}{|p{\dimexpr(\linewidth) / 2}|p{\dimexpr(\linewidth) / 2}|} \hline
\rowfont[c]{\bfseries}
cell & \inlineImage{https://zestedesavoir.com/media/galleries/426/56dc4a1e-416b-4a9d-830d-95b45d58a17a.png} \\ \hline
\end{longtabu}


\begin{longtabu} spread 0pt {|X[-1]|X[-1]|} \hline
\rowfont[c]{\bfseries}
c1 & c2 \\ \hline
\rowfont[l]{}
[...]

PR #355, opened 2019-08-13
rebber-plugins: fix #352 (\par in codes)

artragis:
#352 

# Resolution

- remove the `replace(/\n/g, ' \\par ')` 
- add the fixture given on the ticket to generate a proper snapshot
- run the jest command and observe old snapshots fails
- develop the fix while previous step is true
- lerna run build --scope rebber-plugins
- run zmarkdown tests

# Note

On my local version I've got issues with zmarkdown tests because of iframes.

PR #356, opened 2019-08-14
Transform images to inline in heading, table and gridTable

TitiAlone:
This PR fixes bug #354 with added **new tests**.

# How it works

The first thing that was done was to write tests for failing cases and correct erroneous snapshot for one of the previous tests. I may have added a lot of tests compared to the problem, but a wise old man once told me there was never too much tests 🧙‍♂️.

Then, on the problem itself, I added a new option for the image serializer of `rebber`, which allow to inject a matcher for when the image should be renderer inline. The default matcher keeps the old `paragraph` exception and add a new one for `heading`; also, this matcher is enforced in `gridTable` of rebber-plugins and `table`of rebber.

Well, well, well, anyway, the bug is now solved and tests are now good! Be careful before merging; we may need a rebase because PR #355 deals with the same block of code.


TitiAlone:
Tests will fail I guess for Node 8 on Travis for whatever reason; tell me if it's my fault @vhf .

vhf:
Nice, thanks!

> Update: I do not know why Travis is so drunk, often fails on some PR for no reason and sometimes resolves automagically...

Not automagic, manualmagic: I restarted the build 😉

Issue #294, opened 2018-10-31
remark-download-images: use default black image when 404


artragis:
When we transform our markdown into latex, and some images lead to 404 error, remark-download-images generate a png file but with html content.

Here is an example from @motet-a tutorial: 

```markown
![sprite](https://zestedesavoir.com/static/images/sprite.f651d381c5ba.png)
```
this image is not available. When generating latex, we get this error : 

```json
[
    {
        "column": 1,
        "name": "2142:1-2142:97",
        "line": 2142,
        "reason": "Received HTTP404 for: https://zestedesavoir.com/static/images/sprite.f651d381c5ba.png",
        "ruleId": "//zestedesavoir.com/static/images/sprite.f651d381c5ba.png",
        "fatal": false,
        "stack": "Error: Received HTTP404 for: https://zestedesavoir.com/static/images/sprite.f651d381c5ba.png\n    at Request.request.get.on (/opt/zmd/node_modules/remark-images-download/src/index.js:140:26)\n    at emitOne (events.js:121:20)\n    at Request.emit (events.js:211:7)\n    at Request.onRequestResponse (/opt/zmd/node_modules/request/request.js:1066:10)\n    at emitOne (events.js:116:13)\n    at ClientRequest.emit (events.js:211:7)\n    at HTTPParser.parserOnIncomingClient (_http_client.js:551:21)\n    at HTTPParser.parserOnHeadersComplete (_http_common.js:115:23)\n    at TLSSocket.socketOnData (_http_client.js:440:20)\n    at emitOne (events.js:116:13)\n    at TLSSocket.emit (events.js:211:7)\n    at addChunk (_stream_readable.js:263:12)\n    at readableAddChunk (_stream_readable.js:250:11)\n    at TLSSocket.Readable.push (_stream_readable.js:208:10)\n    at TLSWrap.onread (net.js:607:20)",
        "location": {
            "start": {
                "column": 1,
                "offset": 102942,
                "line": 2142
            },
            "indent": [],
            "end": {
                "column": 97,
                "offset": 103038,
                "line": 2142
            }
        },
        "message": "Received HTTP404 for: https://zestedesavoir.com/static/images/sprite.f651d381c5ba.png",
        "source": "https"
    }
]

```

When image is not properly formatted, latex fails (see https://github.com/zestedesavoir/latex-template/issues/98 ) so I wanted to read the error to generate a fake "black" image to ensure the pdf to be generated.

[...]

PR #357, opened 2019-08-15
Allow to use default black image when server error on image downloading

TitiAlone:
Solves issue #294 in two parts:

## `remark-images-download`: add an optional parameter

A new option was added to the package to allow replacing the default image path with a custom "error image" on three different cases:

- status code matches an error (!= 200);
- MIME type is incorrect;
- file exceed max size.

The first thing to do is to enable the option - please note that **no breaking changes** were made, the package still works as it did before, a new option was simply added - using `defaultImagePath` which corresponds to the path of the image to fallback to. Once this is done, simply tell the plugin when it should fallback using `defaultOn`, specified in the README.

Technically, the fallback is done by adding a boolean to the error returned by the Promise (on `reject`), called `error.replaceWithDefault`, and that will allow us to add more cases for the fallback later if needed. When this boolean is matched to `true`, the AST is mutated by changing the image URL to the concatenated default path.

Three tests were also added for the three use cases independently.

## `zmarkdown` add an option to the LaTeX endpoint

On ZMarkdown side, the change **is possibly breaking**, as it changes the default configuration and could change the HTML rendering if these defaults are used. A new option, `images_download_default` is available for endpoints that allow images download.

By default, fallback is done to `black.png`, but can be changed using EPUB or LaTeX endpoint option, as @artragis requested. Of the three cases mentionned above, only one is used here right now: erroneous status code, because we only need this case for Zeste de Savoir; if another case is to be used in the future, a simple change to `remark.js` configuration file is needed.

Please review @vhf and tell me if it looks good.

### DIFF
--- a/packages/remark-images-download/README.md
+++ b/packages/remark-images-download/README.md
@@ -33,6 +33,12 @@ unified()
   .use(remarkImagesDownload, {
     disabled: true,
     downloadDestination: './img/',
+    defaultImagePath: 'black.png',
+    defaultOn: {
+      statusCode: true,
+      mimeType: false,
+      fileTooBig: false,
+    },
     maxlength: 1000000,
     dirSizeLimit: 10000000,
     localUrlToLocalPath: (localUrl) => localPath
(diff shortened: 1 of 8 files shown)


vhf:
Merged: 795b6a5a6a80ce4ac9b5b9284d6d9f38554cd19e

Issue #358, opened 2019-08-15
[remark-disable-tokenizers] autoLink replaced with noop but still tokenizes


Redmega:
As the title says, autoLink persists through the disable.

See this codesandbox for an example.
https://codesandbox.io/embed/remark-disable-tokenizers-autolink-bug-c9q7b

```
  parseMarkdown = unified()
    .use(parse, { pedantic: false })
    .use(remarkDisableTokenizers, {
      inline: ['autoLink'],
    })
    .use(remarkRehype)
    .use(rehypeReact, {
      Fragment: React.Fragment,
      createElement: React.createElement,
    })
```

TitiAlone:
Hello,

thanks for reporting the problem; for some reasons I'm not aware of, you need to disable the `url` tokenizer for your example to work. Actually, it seems that `autoLink` does nothing at all, links do not even pass through it on `remark-parse`.

As I'm quite sure someone would ask me whether this disables "classic linking", please note that it does not, as another tokenizer, whose name is `link`, handles it separately.

Sorry for the trouble, but we can't do much ourselves, maybe we could map `autoLink` to `url` directly in the source, but I'm quite sure it would cause some trouble later and we better no do it. Maybe @vhf could have more information about this.

Redmega:
Sounds like this might actually be an issue with `remark-parse` then... Thanks for the find @TitiAlone ! At least we have a way forward with this.

PR #360, opened 2019-08-19
[remark-disable-tokenizers] Add caveat about autoLinking to README

Redmega:
As per #358, autoLink currently does nothing for some reason. Until the core issue is found and fixed, others trying to disable autoLinking might find it useful to know they should disable the  `url` tokenizer instead.


vhf:
Nice, thanks!

PR #362, opened 2019-08-24
fix: update labels to use footnote numbers as well as identifiers

### LINKED ISSUE #361: remark-numbered-footnotes should use `string` identifiers
https://github.com/zestedesavoir/zmarkdown/blob/a9cba16b5237218a2b59820bda599ee4b0410351/packages/remark-numbered-footnotes/src/index.js#L37

...should be:

```
 node.identifier = String(footnotes[identifier] )
```

...as mdast [requires it to be a string](https://github.com/syntax-tree/mdast#association)!

jlengstorf:
As of https://github.com/syntax-tree/mdast-util-to-hast/pull/31, the footnote label is used for footnotes instead of its identifier. This change brings back numbered footnotes correctly.

This was identified in this issue: https://github.com/gatsbyjs/gatsby/issues/16578

Once this is merged, we're able to fix the affected Gatsby plugin: https://github.com/jlengstorf/gatsby-remark-numbered-footnotes/issues/9

fixes #361

### DIFF
--- a/packages/remark-numbered-footnotes/src/index.js
+++ b/packages/remark-numbered-footnotes/src/index.js
@@ -34,8 +34,8 @@ function createIds (footnotes) {
     if (!footnotes.hasOwnProperty(identifier)) {
       footnotes[identifier] = Object.keys(footnotes).length + 1
     }
-    node.identifier = footnotes[identifier]
-    node.label = footnotes[identifier]
+    node.identifier = String(footnotes[identifier])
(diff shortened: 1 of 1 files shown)


TitiAlone:
Solving #361 at the same time would be a good idea; would you have time to do it? Resolution is provided in the issue.

jlengstorf:
@TitiAlone done!

Issue #364, opened 2019-09-25
[rebber-plugins/zmarkdown]emoticon node type is not created


artragis:
In zmarkdown, and all projects that uses remark-emoticon, we add a `emoticon` node type on the MDAST.

Neither `rebber-plugins` nor `zmarkdown` references this type to stringify emoticon, therefore produced texfile/pdf do not embed emoticons.

# expectation

for markdown `:)` the latex endpoint in zmarkdown must render `\smiley{smile.png}`.

vhf:
> for markdown `:)` the latex endpoint in zmarkdown must render `\smiley{smile.png}`.

With our configuration, this would be `\smiley{/static/smileys/smile.png}`, correct?

artragis:
My expected code is the right one, this means we do not correctly configure remark-emoticon for rebber : the emoticon path is defined in `\smileyDict` command and then we give the smiley.

vhf:
Could you please give a link to `\smileyDict`?

artragis:
The current is `/opt/zds/app/dist/smileys` but you really don't need it, you just have to push `\smiley{smile.png}` for example.

artragis:
Thanks a lot

PR #365, opened 2019-09-28
Solves #364

TitiAlone:
It seems that the `emoticons` object needed was moved into `emoticons.emoticons` when `class` was added to `rehype-emoticons`.

#364


artragis:
can you add a test that ensures /latex endpoints renders the emoticon please.

vhf:
Thanks!

Issue #345, opened 2019-07-28
Mathml and epub : missing xmlns and xml-annotation


artragis:
When we process math with KaTeX for epub we output something like this : 
```html
<span class="inlineMath">
      <span class="katex">
       <span class="katex-mathml">
        <math>
         <semantics>
          <mrow>
           <mfrac>
            <mrow>
             <mi>
              N
             </mi>
             <mo>
              ×
             </mo>
             <mo>
              (
             </mo>
             <mi>
              N
             </mi>
             <mo>
              +
             </mo>
             <mn>
              1
             </mn>
             <mo>
              )
             </mo>
            </mrow>
            <mn>
             2
            </mn>
           </mfrac>
          </mrow>
          <annotation>
           \frac {N \times (N+1)} {2}
          </annotation>
         </semantics>
        </math>
       </span>
       <span class="katex-html">
        <span class="base">
         <span class="strut" style="height:1.355em;vertical-align:-0.345em;">
         </span>
         <span class="mord">
          <span class="mopen nulldelimiter">
          </span>
          <span class="mfrac">
           <span class="vlist-t vlist-t2">
            <span class="vlist-r">
             <span class="vlist" style="height:1.01em;">
              <span style="top:-2.6550000000000002em;">
               <span class="pstrut" style="height:3em;">
               </span>
               <span class="sizing reset-size6 size3 mtight">
                <span class="mord mtight">
                 <span class="mord mtight">
                  2
                 </span>
                </span>
               </span>
              </span>
              <span style="top:-3.23em;">
               <span class="pstrut" style="height:3em;">
               </span>
               <span class="frac-line" style="border-bottom-width:0.04em;">
               </span>
              </span>
              <span style="top:-3.485em;">
               <span class="pstrut" style="height:3em;">
               </span>
               <span class="sizing reset-size6 size3 mtight">
                <span class="mord mtight">
                 <span class="mord mathdefault mtight" style="margin-right:0.10903em;">
                  N
                 </span>
                 <span class="mbin mtight">
                  ×
                 </span>
                 <span class="mopen mtight">
                  (
                 </span>
                 <span class="mord mathdefault mtight" style="margin-right:0.10903em;">
                  N
                 </span>
                 <span class="mbin mtight">
                  +
                 </span>
                 <span class="mord mtight">
                  1
                 </span>
                 <span class="mclose mtight">
                  )
                 </span>
                </span>
               </span>
              </span>
             </span>
             <span class="vlist-s">
[...]

Issue #340, opened 2019-07-08
Support for linenostart


arobase-che:
Hi \o

`linenostart` was an option used in the previous markdown processor.

Should we implement it ? It's still referenced in the fixtures. [Search for linenostart](https://github.com/zestedesavoir/zmarkdown/search?q=linenostart&unscoped_q=linenostart)

And in the official MD guide ([Rédiger sur ZdS](https://zestedesavoir.com/tutoriels/249/rediger-sur-zds/#d%C3%A9but-de-la-num%C3%A9rotation))

To implement it we need to update remark to version 6. and modify the file [rehype-line-numbers.js](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/zmarkdown/utils/rehype-line-numbers.js) to add something like `style: 'counter-reset: a X;'` where X is based on the attribute `meta` (implemented in version 6 of remark).

💗

vhf:
👍 

See:
> @arobase-che do you want to work on upgrading remark? Anyone else? I probably won't have time but I'd be happy to help.
> 
> We don't need to wait for a remark release containing this bugfix to work on this.
https://github.com/zestedesavoir/zmarkdown/issues/307#issuecomment-510873857

artragis:
to be closed on release, non?

TitiAlone:
I thought this too, but updating only fix for LaTeX.
But with the new version of remark, we are able to use a new syntax for parsing, so this will be fixed soon.

arobase-che:
Still have to write a some code on the side of rehype-line-number I think.

Issue #307, opened 2018-12-24
A blank line should always separate block quotes


artragis:
The minimum markdown used : 

```
> Bonjour !
Source: Une personne.

> Au revoir !
Source: Une autre personne.
```

# result 

```
<figure><blockquote>
<p>Bonjour !
Source: Une personne.</p>
<p>Au revoir !</p>
</blockquote><figcaption>Une autre personne.</figcaption></figure>
```
# expected

```
<figure><blockquote>
<p>Bonjour !</p>
</blockquote><figcaption>Une personne.</figcaption></figure>
<figure><blockquote><p>Au revoir !</p>
</blockquote><figcaption>Une autre personne.</figcaption></figure>
```

vhf:
This is weird, `gfm` should already take care of this.

arobase-che:
Hi \o

I have some time to study that case.

**tl;dr**: The problem is [here](https://github.com/remarkjs/remark/blob/master/packages/remark-parse/lib/tokenizer.js#L307), line 307 of the file tokenizer.js from remark-parse.

It's a remark problem since : 

~~~js
const unified = require('unified');
const markdown = require('remark-parse');
const inspect = require('unist-util-inspect');

const processor = unified()
  .use(markdown, {gfm: true});

const ast = processor.parse(`

> Bonjour !
Source: Une personne.

> Au revoir !
Source: Une autre personne.
`);
console.log(inspect(ast));
~~~

Yield : 
~~~
root[1] (1:1-8:1, 0-79)
└─ blockquote[2] (3:1-7:28, 2-78)
   ├─ paragraph[1] (3:3-4:22, 4-35)
   │  └─ text: "Bonjour !\nSource: Une personne." (3:3-4:22, 4-35)
   └─ paragraph[1] (6:3-7:28, 39-78)
      └─ text: "Au revoir !\nSource: Une autre personne." (6:3-7:28, 39-78)
~~~

Should I open an issue and submit a PR in remark referencing [gfm#213](https://github.github.com/gfm/#example-213) ?

artragis:
I think you should, yes.

arobase-che:
Ok done. remark/remarkjs#415

It is gfm#220 not gfm#213 btw.

arobase-che:
Fixed in remarkjs/remark ! ^^

We need to update remark to take into account the fix :)  
But we have to wait for a next version :(

vhf:
Nice, thanks!
[...]

Issue #255, opened 2018-07-24
hl_lines is not supported anymore


artragis:
With zmarkdown, when we want to use hl_lines on source codes, the render to rehype is not properly parsed.

# the markdown

# what we get
```html
        <div class="hljs-code-div"><div class="hljs-line-numbers"><span></span><span></span><span></span><span></span></div><pre><code class="hljs language-python">a = <span class="hljs-number">2</span>
b = a + <span class="hljs-number">5</span>
c = a + b
print(a, b, c)
</code></pre></div>
```
# what we expect : 

something that wraps line 3 (and all lines that are specified) so that css can color it. For example, it can be

```html
        <div class="hljs-code-div"><div class="hljs-line-numbers"><span></span><span></span><span></span><span></span></div><pre><code class="hljs language-python">a = <span class="hljs-number">2</span>
b = a + <span class="hljs-number">5</span>
<span class="hljs-highlight">c = a + b</span>
print(a, b, c)
</code></pre></div>
```

Nota : here are the supported syntax to specify lines : 

```javascript
hl_lines=3 // only line 3 will be highlighted
hl_lines=3,4 // line 3 and line 4
hl_lines=3-8 // line 3 to 8
hl_lines=3,5-8,15 // line 3, lines 5 to 8, line 15 will be highlighted
```

Heziode:
So, line highlight is not implemented and it should not be implemented in highlight.js (see highlightjs/highlight.js#740). Remark use highlight.js for code highlightning.

We need to implement a wrapper to support this feature.

TRSasasusu:
Today, I made a plugin highlighting specific lines.
https://github.com/TRSasasusu/highlightjs-highlight-lines.js

artragis:
Thanks @TRSasasusu. It seems that your plugin requires a browser to run.

TRSasasusu:
Oops, I understood...
So, to make this plugin work in this project, this plugin should work in [lowlight](https://github.com/wooorm/lowlight) used in rehype.

However, I'm sorry, but I'm tied up now, so please wait for dealing with it by me (or please could you do it...?)

TitiAlone:
Since this is now done, I think it can be closed!

PR #371, opened 2020-01-30
Update dependencies

TitiAlone:
This is a retake on @arobase-che's previous work on updating dependencies for zmarkdown packages.

## List of changed behaviours

Here's a list of what has changed (and hence, what needs to be tested on zds-site) for reference, made by looking at the modified snapshots and fixtures:

- a blank line now separate blockquotes (solves #307);
- use \footnote instead of list of links for references in LaTeX;
- empty code spans do not render anything anymore;
- KaTeX SVG has the `xmlns` attribute (solves one half of #345);
- lists with empty items render a list instead of nothing;
- tables now take precedence over code blocks;
- footnotes are wrapped into a paragraph when outside a list;
- footnotes are not anymore wrapped into a paragraph when part of a list;
- footnotes order has changed ([see below](https://github.com/zestedesavoir/zmarkdown/pull/371#discussion_r373767234));
- generally, paragraphs are not added inside lists anymore;
- hl_lines and linenostart parameters for code blocks are back again **for LaTeX only** (solves one half #255 and #340);
- captions now break lists (this was expected by a test, which was incorrect before);
- quite a lot of changes are expected from KaTeX, please report any issue found;
- by the way, KaTeX was updated to version 0.11.1, so please update KaTeX also on zds-site (to 0.11.1) when deploying this version of zmarkdown.

## What remains?

First of all, I would like a global approval of the changes above by @artragis .
Maybe a review is needed (poke @vhf ), but I honestly don't know how this PR could be checked, maybe by looking at the updated tests.
Most importantly, it would have been great to do a rebase, but this is very complicated to do correctly here, so I think it will be a simple merge.
Finally, dependencies of `remark-image-download` were NOT updated, because they would need huge refactoring of the code, and I prefer to merge this first before attempting to do it (changes are less urgent, because they don't concern security issues nor are they feature-related).


TitiAlone:
Now with rebase, please check but I think it is okay; I had to do a commit because some snapshots were still in their old version after the rebase.

packages/rehype-postfix-footnote-anchors/__tests__/__snapshots__/index.js.snap
```diff
@@ -5,18 +5,9 @@ exports[`{"gfm":false,"commonmark":false,"footnotes":true} footnote-split 1`] =
...
-<p>third def</p>
-<a href=\\"#fnref-third-postfix\\" class=\\"footnote-backref\\">↩</a>
-</li>
-<li id=\\"fn-1-postfix\\">
-<p>second def</p>
-<a href=\\"#fnref-1-postfix\\" class=\\"footnote-backref\\">↩</a>
-</li>
+<li id=\\"fn-first-postfix\\">first def<a href=\\"#fnref-first-postfix\\" class=\\"footnote-backref\\">↩</a></li>
+<li id=\\"fn-1-postfix\\">second def<a href=\\"#fnref-1-postfix\\" class=\\"footnote-backref\\">↩</a></li>
+<li id=\\"fn-third-postfix\\">third def<a href=\\"#fnref-third-postfix\\" class=\\"footnote-backref\\">↩</a></li>
```
vhf: Corresponding fixture:

```md
a[^first]b[^second def]c[^third]

[^first]: first def
[^third]: third def
```

The behavior changed, probably for the better.

packages/remark-captions/__tests__/__snapshots__/index.js.snap
```diff
@@ -21,9 +21,10 @@ print('bla')
 exports[`captions should break lists 1`] = `
 "<figure><blockquote>
 <ul>
-<li>Bar
-Source: fc2</li>
+<li>Bar</li>
 </ul>
+</blockquote><figcaption>fc2</figcaption></figure>
+<figure><blockquote>
```
vhf: Nice side effect of the new "blank lines between quotes create separate quotes" behavior.

packages/remark-grid-tables/package.json
```diff
@@ -33,8 +33,9 @@
   "license": "MIT",
   "dependencies": {
     "grapheme-splitter": "^1.0.4",
+    "is-fullwidth-code-point": "^3.0.0",
```
vhf: Extraneous dep: not actually used in this package. That's just a tiny merge mistake.

packages/remark-images-download/src/index.js
```diff
@@ -1,3 +1,6 @@
+require('core-js/stable')
+require('regenerator-runtime/runtime')
```
vhf: Both of these seem overkill. Upgrading this package independently as you suggested makes sense IMO, I'd suggest reverting this file and only bumping unist-util-visit/rimraf/shortid for now.

packages/zmarkdown/__tests__/__snapshots__/latex.tests.js.snap
```diff
@@ -285,7 +290,7 @@ exports[`list 1`] = `
 `;
 
 exports[`math 1`] = `
-"A sentence ($S$) with \\\\textit{italic} and inline math ($C_L$) and $$b$$ another.
+"A sentence ($S$) with \\\\textit{italic} and inline math ($C_L$) and $b$ another.
```
vhf: Not sure where this change comes from and if it's intentional.

packages/zmarkdown/__tests__/new-html-suite.test.js
```diff
@@ -137,7 +137,7 @@ describe('tables', () => {
       \`ici ok\`| \`gauche | droite\`
     `
 
-    return expect(renderString(input)).resolves.toContain('<td><code>gauche | droite</code></td>')
+    return expect(renderString(input)).resolves.toContain('<td>`gauche</td>')
```
vhf: I don't think that's correct, the last cell should contain `<code>gauche | droite</code>`:

```md
Titre 1 | Titre 2
--------|--------
`ici ok`| `gauche | droite`
```

TitiAlone: I thought the same, so I opened an issue on [remark](https://github.com/remarkjs/remark/issues/465), and this is actually the expected behaviour.

packages/rehype-postfix-footnote-anchors/__tests__/__snapshots__/index.js.snap
TitiAlone: I don't know what you think changed, but from what I can see, only the order of the three footnotes changed.

packages/remark-captions/__tests__/__snapshots__/index.js.snap
TitiAlone: I had a doubt on this one, but the title "captions should break lists" led me into thinking that the current behaviour was the expected one.

packages/rehype-postfix-footnote-anchors/__tests__/__snapshots__/index.js.snap
vhf: Yeah I was talking about the order in which footnotes are rendered. Footnotes are displayed in the same order as their definition.

```md
a[^first_footnote_reference]

b[^`b` second footnote reference but first footnote definition]

c[^last_footnote_reference]

[^last_footnote_reference]: `c` last footnote reference but second footnote definition
[^first_footnote_reference]: `a` first footnote reference but last footnote definition
```

Before:

image

After:

image

In the HTML ordered list, the list items 1/2/3 don't match the footnote references numbers.

TitiAlone: Ok, I see what you mean, added it to the issue status to get an approval of the change.

artragis: this cannot be good : if you click on "3" you get redirected to "c" whereas it should lead you to "a"

vhf: Yeah, the footnote reference (superscript) and the list item number should match.

a^3, 3. a

packages/zmarkdown/__tests__/__snapshots__/latex.tests.js.snap
TitiAlone: You're right, this is a bad behaviour, finally found it and solved it!

packages/zmarkdown/config/remark.js
```diff
@@ -317,10 +317,6 @@ const remarkConfig = {
     inlineMathDouble: true,
   },
 
-  katex: {
-    inlineMathDoubleDisplay: true,
```
artragis: pourquoi retirer celui-là?

TitiAlone: It doesn't exist anymore, the behaviour targeted by this option is now the default; rehype-katex decided to use remark-math's `inlineMathDouble` option (indirectly) to detect the option instead.

packages/rehype-postfix-footnote-anchors/__tests__/__snapshots__/index.js.snap
TitiAlone: This is fixed in zmarkdown; the behaviour is normal in rehype, so the test will not be updated for the package.

artragis:
> captions now break lists (this was expected by a test, which was incorrect before);

can you explain this, And why not openning an issue that will be closed by this PR so that we can track the change?

> by the way, KaTeX was updated to version 0.11.1, so please update KaTeX also on zds-site (to 0.11.1) when deploying this version of zmarkdown.

Having a look at https://github.com/KaTeX/KaTeX/releases, I see 

> Fix \sqrt SVG path

so @Situphen we will clearly need to update katex dependency on zds-site side

TitiAlone:
> can you explain this, And why not openning an issue that will be closed by this PR so that we can track the change?

Please refer to [this test](https://github.com/zestedesavoir/zmarkdown/pull/371/files#diff-b029b179a6eb4d9d5394b1e79f925735). The new behaviour is expected, not a bug, but could maybe be update if it doesn't suit zds-site needs.

> so @Situphen we will clearly need to update katex dependency on zds-site side

Yes.

vhf:
> > captions now break lists (this was expected by a test, which was incorrect before);
> 
> can you explain this

This [used to](https://zestedesavoir.github.io/zmarkdown/) generate a single quote:
```md
> Foo

> Bar
```

Now that it generates two quotes, this should be two quotes and their respective captions:
quote:
```md
> Foo
Source: foo

> Bar
Source: foo
```

Same goes for:
```md
> * Foo
Source: foo

> * Bar
Source: foo
```

artragis:
Ok so everything seems to be fine for me.

Issue #359, opened 2019-08-19
KaTeX: add mchem


pierre-24:
[mhchem](https://github.com/KaTeX/KaTeX/tree/master/contrib/mhchem) (an extension to support chemical equation in KaTeX) is supported [since v0.10.1](https://github.com/KaTeX/KaTeX/pull/1436) in KaTeX (the extension is part of the official release, but does not seems activated by default, or at least not on ZdS), and I have to say I would be very happy if we could include it, since I am a huge fan of it. Concerning rebber, we just need to add the "real" `mhchem` package to the template and that's it.

Many thanks in advance :)

TitiAlone:
Depends on [upstream](https://github.com/remarkjs/remark-math/pull/34); we will be able to implement it right after.

pierre-24:
Nice, thanks :)

PR #376, opened 2020-02-03
Add mhchem extension to KaTeX (solves #359)

TitiAlone:
Add the mhchem extension by using KaTeX globals; also adds to tests to make sure:

1. it is well rendered by KaTeX (avoid version conflicts): the test detects the lack of "KaTeX parse error: Undefined control sequence" message;
2. it is directly used without modifications in rebber (which may be overkill, because it is basically math).

### DIFF
--- a/packages/zmarkdown/__tests__/latex.tests.js
+++ b/packages/zmarkdown/__tests__/latex.tests.js
@@ -193,3 +193,10 @@ test('regression: code block without language', () => {
 
   return expect(p).resolves.toMatchSnapshot()
 })
+
+test('properly loads extensions - mhchem', () => {
+  const markdown = '$\\ce{H2O}$'
+  const result = renderString()(markdown)
+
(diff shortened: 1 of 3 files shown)


artragis:
ping @pierre-24

pierre-24:
Oh <3

Point 2 is OK: the command is basically translated from the corresponding LaTeX package, so it is natural to have it in rebber (it requires the corresponding package, though, see https://github.com/zestedesavoir/latex-template/issues/112

... Either you merge it first and we adapt in the template or the other way around (chicken and egg situation).

TitiAlone:
@pierre-24 : I think it is safer to add it first to `latex-template`, as we are sure adding a package won't break anything, while it may be breaking to add it to zmarkdown while not in `latex-template`.

TitiAlone:
Will merge now, but publish later, so that you have time to adapt the template.

PR #377, opened 2020-02-03
rebber-plugins/zmarkdown: Improve grid table latex rendering

### LINKED ISSUE #374: rebber: array improvements
The goal is to be able to generate LaTeX code with paragraph in multirow. For that, we have to

- replace `\n` by `\endgraf` ([for now, we replace them by `\par`](https://github.com/zestedesavoir/zmarkdown/blob/a6fc38b3c2a1da34e4916c5ccae601f442396df8/packages/rebber-plugins/src/type/gridTable.js#L54)),
- use `\parbox{\linewidth}{text}` in `\multirow` instead of just `text`,
- use `{\centering text}` and `{\raggedleft text}` instead of `\begin{center} text \end{center}` and `\begin{flushright} text \end{flushright}`.

(We can also replace `p{dim}` by `m{dim}` (see [here](https://tex.stackexchange.com/questions/35293/p-m-and-b-columns-in-tables) for the differences)).

artragis:
fix #374 

- uses the "switch command" version of centering and flushright https://cs.overleaf.com/learn/latex/Text_alignment#Right-justified_text
- use parbox instead of nothing in multirow and multicol
- uses `m{dim}` instead of `p{dim}`
- uses  `\endgraph` insteand of `\par` to embody line-break char

### DIFF
--- a/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
+++ b/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
@@ -781,27 +781,24 @@ exports[`mix-3 1`] = `
 
 
 
-\\\\begin{flushright}
+{\\\\raggedleft
 \\\\begin{itemize}
 \\\\item item
 \\\\item item
 \\\\end{itemize}
-
-\\\\end{flushright}
+}
 
 
 \\\\part{centered list}
 
 
 
-\\\\begin{center}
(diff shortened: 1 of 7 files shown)


artragis:
I removed the package-json.lock

TitiAlone:
Merged, thanks; it will be available with the new version of zmarkdown that will get released soon.

Issue #375, opened 2020-02-03
rebber: array issue


Karnaj:
```
+----+----+
| aa | aa |
+====+====+
| aa | aa |
+----+----+
| aa | aa |
+----+----+
```
This code products this LaTeX.
```latex
\begin{longtabu}{|p{\dimexpr(\linewidth) / 2}|p{\dimexpr(\linewidth) / 2}|} \hline
\rowfont[c]{\bfseries}
aa & aa \\ \hline
\rowfont[c]{\bfseries}
aa & aa \\ \hline
\rowfont[l]{}
aa & aa \\ \hline
\end{longtabu}
```
It should product this one.
```latex
\begin{longtabu}{|p{\dimexpr(\linewidth) / 2}|p{\dimexpr(\linewidth) / 2}|} \hline
\rowfont[c]{\bfseries}
aa & aa \\ \hline
\rowfont[l]{}
aa & aa \\ \hline
aa & aa \\ \hline
\end{longtabu}
```

Issue #313, opened 2019-02-13
Forbidden images in Signature (message footer) 


A-312:
zestedesavoir/zds-site/issues/5086

artragis:
In fact, what we want is to be able to add a custom element to the ignored ones when we request zmarkdown endpoints.

@A-312 can you edit the issue to write a well formatted ticket?

TitiAlone:
Added an option for this today, but I'm not sure it will solve your problem, unfortunately; images are tokenized at the same time as links/url in `remark`, which means we cannot disable only one at a time: if we loose images, we loose links...

Issue #283, opened 2018-09-06
[rebber&zmarkdown] add additional options to latex-document enpoint


pierre-24:
It would be interesting to provide additional options to use the following commands, which go in the preamble (before `\begin{document}`):

+  `\logo` and `\editorLogo` (example of use [showed here](https://github.com/zestedesavoir/latex-template/blob/master/tests/test.tex#L9-L10)): `\logo` is the logo of the content, while `\editorLogo` is the logo of the website (as shown in the example, one can also provide a ~~slogan~~ motto with it). I'm not against `*_directory` options.
+ `\tutoLink`: that option should expend to `\tutoLink{${tutolink}}` (link to the content). It is the link around the `\logo` on the title page
+ `\editor`: that should expend to `\editor{${editorlink}}` (link to the editor website). It is the link around `\editorLogo` on the title page.

Obviously, none of those options should be mandatory (it works well without it right now). Thanks in advance :)

PR #380, opened 2020-02-06
[rebber] Lots of corrections and improvements

TitiAlone:
These are the last changes until next release; I plan on publishing a major version of rebber, which requires the following to be solved:

- add `mhchem` (#359): done and merged;
- array improvements (#374): done and merged;
- array issue (#375): done and merged;
- avoid spoiler in spoiler (#280): done, first in this PR;
- allow to disable custom tokenizers (for #313): done, second in this PR but not sure it solves the issue;
- add options to zmarkdown's LaTeX endpoint (#283 *only*): done, third in this PR;
- send codeblocks to appendices (#299): done and merged;
- add a `\ping` command (none): done, fourth in this PR;
- add missing dependencies to rebber (#384): done, fifth in this PR.

We will be able to publish soon enough, now everything is done.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14682,8 +14682,10 @@
     "rebber": {
       "version": "file:packages/rebber",
       "requires": {
+        "clone": "^2.1.2",
         "has": "^1.0.3",
         "mdast-util-definitions": "^2.0.0",
+        "trim-lines": "^1.0.0",
         "xtend": "^4.0.2"
       },
       "dependencies": {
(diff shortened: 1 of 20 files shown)


artragis:
We could add some changelogs.

- in rebber:
    - add missing dependency
    - fix linkReference default macro (uses hyperref from hypertext package instead of bare ref) 
- in rebber-plugins : 
    - fix grid-table/align interraction
    - add ping command
    - fix appendix management
    - avoid imbrication of spoilers
- in zmarkdown:
    - push codes that were inside footnote into appendix
    - add parameter in endpoints

PR #382, opened 2020-02-08
Solve the footnote order problem

TitiAlone:
Really simple fix to the footnote order problem that wasn't really solved when updating dependencies.

Before, we did:

- visit the definitions and number the footnotes in this order;
- apply the numbers to the references.

But recently, `mdast-util-to-hast` broke the module by numbering the footnotes at the end in the references order; until now, we had a trick on zmarkdown to keep the old behaviour, this PR aims to fix the problem directly inside the package:

- visit the references and number the footnotes in this order;
- apply the numbers to the definitions.

This behaviour needs to be validated by @artragis though, as it does differ from the previous one: when updating a content, the footnotes order could change.

### DIFF
--- a/packages/remark-numbered-footnotes/README.md
+++ b/packages/remark-numbered-footnotes/README.md
@@ -1,37 +1,5 @@
-This plugin replaces the footnote references with a number sequence (starting from 1) in the same order as the footnote definitions (**not** the footnote references).
-
-Reordering the definitions (usually put at the end of the document in the Markdown source) will therefore let you reorder the sequence.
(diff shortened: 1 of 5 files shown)


TitiAlone:
What do you think about keeping the changes, and reordering only in zmarkdown for our specific case?

vhf:
My worry is not about how zmd behaves, it's about this package changing in a way that breaks backward compatibility. It could be made configurable though, I don't mind as long as people can use this package to number footnotes based on definition order.

TitiAlone:
I think this is great for us: https://github.com/remarkjs/remark-footnotes.

Footnotes are now handled by a specific package; therefore, I close this issue, and it will need to be adressed in the next dependencies update.

PR #383, opened 2020-02-14
rebber/rebber-plugins: fix appendix management for gridtables and footnotes

artragis:
Hello,

I tried to fix the "code in footnotes" issue and found some annoying issues in the way we manage appendices.

- references were totally wrong
- no links were provided
- many heading "appendix"
- same appendix number if multi table

This is the state of my work, I still need to know:

- How can I say I am inside an appendix?
- Which command is good to link to reference?
- is there a better way to do what we do?

### DIFF
--- a/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
+++ b/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
@@ -325,7 +325,7 @@ space & \\\\inlineImage{https://i.ytimg.com/vi/lt0WQ8JzLz4/maxresdefault.jpg} \\
 title & code \\\\\\\\ \\\\hline
 \\\\rowfont[l]{}
 inline & \\\\texttt{inline} br \\\\endgraf \\\\texttt{inline} \\\\\\\\ \\\\hline
-block & \\\\hyperref[appendix-1]{code} \\\\\\\\ \\\\hline
+block & \\\\hyperref[appendix-1]{Code appendice 1} \\\\\\\\ \\\\hline
 \\\\end{longtabu}
 
 
@@ -456,8 +456,6 @@ elem & \\\\multicolumn{2}{|m{\\\\dimexpr(\\\\linewidth) * 2 / 5}|}{\\\\parbox{\\
 \\\\begin{CodeBlock}{python}
 print('bla')
 \\\\end{CodeBlock}
-
-
 \\\\end{appendices}"
 `;
 
@@ -836,7 +834,7 @@ space & \\\\inlineImage{https://i.ytimg.com/vi/lt0WQ8JzLz4/maxresdefault.jpg} \\
 title & code \\\\\\\\ \\\\hline
 \\\\rowfont[l]{}
 inline & \\\\texttt{inline} br \\\\endgraf \\\\texttt{inline} \\\\\\\\ \\\\hline
-block & \\\\hyperref[appendix-1]{code} \\\\\\\\ \\\\hline
+block & \\\\hyperref[appendix-1]{Code appendice 1} \\\\\\\\ \\\\hline
 \\\\end{longtabu}
 
 
@@ -959,8 +957,6 @@ b &  &  & c \\\\\\\\ \\\\hline
 \\\\begin{CodeBlock}{python}
 print('bla')
 \\\\end{CodeBlock}
-
-
 \\\\end{appendices}"
 `;
 
@@ -990,7 +986,7 @@ exports[`mix-7 1`] = `
 \\\\rowfont[c]{\\\\bfseries}
 cell & \\\\texttt{code with line break} \\\\\\\\ \\\\hline
 \\\\rowfont[l]{}
-\\\\multicolumn{2}{|m{\\\\dimexpr(\\\\linewidth) * 2 / 2}|}{\\\\parbox{\\\\linewidth}{\\\\hyperref[appendix-1]{code}}} \\\\\\\\ \\\\hline
+\\\\multicolumn{2}{|m{\\\\dimexpr(\\\\linewidth) * 2 / 2}|}{\\\\parbox{\\\\linewidth}{\\\\hyperref[appendix-1]{Code appendice 1}}} \\\\\\\\ \\\\hline
 \\\\end{longtabu}
 
 
@@ -1001,8 +997,6 @@ cell & \\\\texttt{code with line break} \\\\\\\\ \\\\hline
 line 1
 line2
 \\\\end{CodeBlock}
-
-
 \\\\end{appendices}"
 `;
 
--- a/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
+++ b/packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
@@ -451,7 +451,7 @@ elem & \\\\multicolumn{2}{|m{\\\\dimexpr(\\\\linewidth) * 2 / 5}|}{\\\\parbox{\\
 
 
 \\\\begin{appendices}
-\\\\label{appendix-1}appendix 1
+\\\\part{\\\\label{appendix-1}Annexes 1}
 
 \\\\begin{CodeBlock}{python}
 print('bla')
@@ -954,7 +954,7 @@ b &  &  & c \\\\\\\\ \\\\hline
 
 
 \\\\begin{appendices}
-\\\\label{appendix-1}appendix 1
+\\\\part{\\\\label{appendix-1}Annexes 1}
 
 \\\\begin{CodeBlock}{python}
 print('bla')
@@ -995,7 +995,7 @@ cell & \\\\texttt{code with line break} \\\\\\\\ \\\\hline
 
 
 \\\\begin{appendices}
-\\\\label{appendix-1}appendix 1
+\\\\part{\\\\label{appendix-1}Annexes 1}
 
 \\\\begin{CodeBlock}{text}
 line 1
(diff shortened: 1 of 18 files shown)


packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
...
-            identifier: `appendix-${appendixIndex}`,
-            referenceType: 'full',
-            children: [{type: 'text', value: 'code'}],
+            type: 'heading',
+            children: [
+
+              {
+                type: 'definition',
+                identifier: `appendix-${appendixIndex}`,
+                url: `${ctx.codeAppendiceTitle || 'Appendix'} ${appendixIndex}`,
```
TitiAlone: Is `appendixIndex` really different from `index` here? I'm thinking these two should be equal according to the way you count.

packages/zmarkdown/__tests__/__snapshots__/old-suite-latex.test.js.snap
```diff
@@ -21,11 +21,11 @@ This \\\\& that.
 
 
 
-Here's a link\\\\ref{1} with an ampersand in the URL.
+Here's a \\\\href{1}{link} with an ampersand in the URL.
```
TitiAlone: Two comments here:

1. I'm not sure this is the expected behaviour, you're basically putting the hyperlink on the word `link` rather of the `1` above that will be suppressed;
2. as far as I know, the `\href` command is not used to refer to the current document in LaTeX but rather intended for external links; maybe you're looking for `\hyperref`, but please check 1. first.

packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
...
+      appendix = {
+        type: 'appendix',
+        children: [],
+      }
+      tree.children.push(appendix)
+    }
+    let appendixIndex = appendix.children.length + 1
     visit(node, 'code', (innerNode, index, _parent) => {
-      inAppendix.push({
+
```
TitiAlone: Extra newline

packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
...
+                children: [{type: 'text', value: 'code'}],
+              },
+            ],
+            depth: 1,
           },
-          {type: 'text', value: '\n'},
-          innerNode,
         ],
       })
+      appendix.children.push(innerNode)
```
TitiAlone: Missing newlines (before & after would be great)

packages/rebber/src/types/linkReference.js
```diff
@@ -1,6 +1,6 @@
 module.exports = linkReference
 
-const defaultMacro = (reference, inner) => `${inner}\\ref{${reference}}`
+const defaultMacro = (reference, inner) => `\\href{${reference}}{${inner}}`
```
TitiAlone: These two are not the same in LaTeX, as pointed below

packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
 const visit = require('unist-util-visit')
+const appendix = require('../type/appendix')
 
 module.exports = (ctx, tree) => {
-  const title = ctx.codeAppendiceTitle || 'Appendices'
+  ctx.overrides.appendix = appendix
```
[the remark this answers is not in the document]
artragis: appendix is not a a mapping from a remark-plugin this is something really specific to preprocessors of rebber-plugins and will only be used when codeBlockVisitor is used

packages/zmarkdown/__tests__/__snapshots__/old-suite-latex.test.js.snap
artragis: `\ref` is not an hyperlink, or we need to redefine `\ref` on latex side.

TitiAlone: Not sure what you mean here, but I'm quite sure that:

- `\ref` is a link to the current document;
- `\href` is a link to an external ressource.

packages/rebber/src/types/linkReference.js
artragis: yep and `\ref` was not really the good command imho.

packages/rebber-plugins/src/preprocessors/codeVisitor.js
artragis: I don't understant what youare meaning here.

index is the index of current visited not in its parent. If the parent is a table cell, index will almost always be 0. Whereas we need the appendixIndex to be incremented each time we add a code into it

packages/zmarkdown/__tests__/__snapshots__/old-suite-latex.test.js.snap
artragis: - `\ref` is not a link, it's just a way to latex to print the reference in the format we asked (e.g figure 1 or Part 1) 
- `\href`is a link whatever the target, it's from hypertext package
- `externalLink` is a link to an external target, I don't know the package

As seen with @karnaj we will use `\hyperref[label]{printed text}`

packages/rebber-plugins/dist/preprocessors/codeVisitor.js
```diff
@@ -2,49 +2,53 @@
 
 var visit = require('unist-util-visit');
 
+var appendix = require('../type/appendix');
```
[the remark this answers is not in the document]
artragis: in fact I just forgot to add it to git.

packages/rebber-plugins/__tests__/__snapshots__/rebber.test.js.snap
```diff
@@ -450,13 +450,15 @@ elem & \\\\multicolumn{2}{|m{\\\\dimexpr(\\\\linewidth) * 2 / 5}|}{\\\\parbox{\\
...
+\\\\begin{appendices}
+\\\\part{\\\\label{appendix-1}Annexes 1}
 
-
-\\\\footnote{\\\\label{appendix-1}}
 \\\\begin{CodeBlock}{python}
 print('bla')
-\\\\end{CodeBlock}"
+\\\\end{CodeBlock}
+
```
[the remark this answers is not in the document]
artragis: I don't know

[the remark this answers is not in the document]
TitiAlone: Comes from the paragraph node, see the other comment.

packages/rebber-plugins/src/type/appendix.js
```diff
@@ -0,0 +1,20 @@
...
+}
+
+function appendixPlugin (ctx, node) {
+  if (node.children.length === 0) {
+    return ''
+  }
+  const overriddenCtx = clone(ctx)
+  overriddenCtx.definition = definitionMacro
+  const innerText = all(overriddenCtx, node)
+  return `\\begin{appendices}\n${innerText}\n\\end{appendices}`
```
TitiAlone: Please do:

```diff
-  return `\\begin{appendices}\n${innerText}\n\\end{appendices}`
+  return `\\begin{appendices}\n${innerText.slice(0,-2)}\n\\end{appendices}`
```

To prevent newlines at the end of the appendices.

packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
...
+
+      appendix.children.push(innerNode)
+
       const referenceNode = {
         type: 'linkReference',
         identifier: `appendix-${appendixIndex}`,
+        children: [
+          {
+            type: 'text',
+            value: 'code',
```
TitiAlone: Please give another name than `code` because it is awfully unclear, maybe `Code : annexe ${appendixIndex}` is good.

packages/rebber-plugins/src/preprocessors/codeVisitor.js
```diff
@@ -1,41 +1,53 @@
...
-            referenceType: 'full',
-            children: [{type: 'text', value: 'code'}],
+            type: 'heading',
+            children: [
+              {
+                type: 'definition',
+                identifier: `appendix-${appendixIndex}`,
+                url: `${ctx.codeAppendiceTitle || 'Appendix'} ${appendixIndex}`,
+                referenceType: 'full',
+                children: [{type: 'text', value: 'code'}],
```
TitiAlone: I didn't noticed this children at the first time, why is it here, I don't think it is needed. It's not even parsed it seems...

packages/rebber-plugins/src/type/appendix.js
vhf: Sounds a bit dangerous if you only want to trim, use `String.prototype.trimEnd()` instead?

TitiAlone: I like it; updated my comment.

packages/rebber-plugins/src/preprocessors/codeVisitor.js
artragis: Oh yes, this is the rest of the old code

packages/rebber-plugins/src/preprocessors/codeVisitor.js
artragis: I was wondering : we could add another parameter "codeAppendinxLabel" to change this. This will avoid putting hard coded french word inside rebber-plugins module

TitiAlone: I'm fine with this option.

artragis: Done for me

packages/rebber-plugins/src/preprocessors/codeVisitor.js
artragis: removed

packages/rebber-plugins/src/type/appendix.js
artragis: updated, thanks for the reviex

artragis:
hum do you know why travis failed?

TitiAlone:
I do, and it's why I merged it anyway, it is solved by now; the test was updated (most likely) by mistake, because the expected behaviour is an iframe (see [here](https://github.com/zestedesavoir/zmarkdown/blob/b7c4b6121182a1526f9b5c8aa0f36065827f609b/packages/zmarkdown/__tests__/__snapshots__/latex.tests.js.snap)).

PR #388, opened 2020-03-11
zmarkdown/rebber-plugins/mdast-split-by-heading: uses introduction commands in latex

artragis:
Here is my PR to implement custom introduction/conclusion commands inside latex document.

This is a WIP because I have some difficulties to run the tests.

Moreover I don't know if useing `\levelOneIntroduction` is good as introductions may have blocks, headings... @karnaj might have a say about it.


TitiAlone:
`splitted` is oftentimes empty, you should handle this case in `introductionAndConclusionVisitor.js`:

```diff
     const splitted = splitAtDepth(rootNode, currentDepth + 1)
+    if(!splitted || !splitted.subTrees) return (node) => {}
     const conclusionCandidate = splitted.subTrees[splitted.subTrees.length]
```

It should allow the tests to pass, but anyway I'm not convinced with your code right now, will have to take a closer look.

packages/rebber-plugins/src/type/introduction.js
```diff
@@ -0,0 +1,9 @@
+module.exports = introduction
+const all = require('rebber/dist/all')
+const introductionCommands = ['levelOneIntroduction', 'levelTwoIntroduction',
```
Karnaj: `Level`, not `level`.

artragis:
your fix does not change anything.

TitiAlone:
Solves all failing tests on my local checkout, maybe you forgot to build? What's for sure is that you didn't push the changes.

artragis:
Ok, the problem was I did not returned (node) => {} even if I don't understand such a return

packages/rebber/dist/escaper.js
```diff
@@ -22,9 +22,8 @@ var defaultEscapes = {
   '{': '\\{',
   '}': '\\}',
   '~': '\\textasciitilde{}'
-  /* Encode special characters in `value`. */
-
 };
+/* Encode special characters in `value`. */
```
TitiAlone: No changes here, please

packages/remark-custom-blocks/dist/index.js
```diff
@@ -122,11 +122,11 @@ module.exports = function blockPlugin() {
     var blockChildren = [contents];
 
     if (titleAllowed && blockTitle) {
-      var configuredTitleBlockType = potentialBlock.titleElement ? potentialBlock.titleElement : 'div';
+      var titleElement = potentialBlock.titleElement ? potentialBlock.titleElement : 'div';
```
TitiAlone: Why is this here? Wasn't it merged with the other PR?

packages/zmarkdown/server/handlers.js
```diff
@@ -162,6 +164,14 @@ module.exports = function markdownHandlers (Raven) {
         ])
       }
 
+      if (opts.content_type === 'big') {
+        rebber.preprocessors.root = introductionAndConclusionVisitor(2)
+      } else if (opts.content_type === 'middle') {
+        rebber.preprocessors.root = introductionAndConclusionVisitor(1)
+      } else {
+        rebber.preprocessors.root = introductionAndConclusionVisitor(0)
+      }
```
TitiAlone: I feel like having named arguments would be way easier to understand.

packages/rebber-plugins/src/preprocessors/introductionAndConclusionVisitor.js
```diff
@@ -0,0 +1,27 @@
+const splitAtDepth = require('mdast-util-split-by-heading/dist')
+
+function introductionAndConclusionVisitor (headingDepth = 0, currentDepth = 0) {
+  return (ctx, rootNode) => {
+    const splitted = splitAtDepth(rootNode, currentDepth + 1)
+    const conclusionCandidate = splitted.subTrees[splitted.subTrees.length]
```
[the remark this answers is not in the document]
artragis: Nice catch !

packages/rebber/dist/escaper.js
artragis: grr this is learna build that creates that.

TitiAlone: You can run it locally using `--scope=rebber-plugins`

artragis: As I needed many plugins, I did not use scope but this is really annoying if the comman dresult is not consistant accross computers.

Issue #400, opened 2020-05-12
rebber: footenotetext cannot be placed everywhere


pierre-24:
This code actually works after translation in LaTeX:

```md
# The elephant [^a]
[^a]: in the room

!![](https://www.drawingforall.net/wp-content/uploads/2015/03/4-apple-drawing.jpg)
Figure: this is a famous [^b] picture
[^b]: or not
```

This one don't

```md
# The elephant [^ in the room]

!![](https://www.drawingforall.net/wp-content/uploads/2015/03/4-apple-drawing.jpg)
Figure: this is a famous [^ or not] picture
```

-------

I'm a big fan of the `[^ this is a footnote]` feature, which is translated by rebber as

```latex
\textsuperscript{\ref{footnote:2}}\footnotetext[2]{\label{footnote:2} this is a footnote.}
```

Except I just discovered that this is problematic in (at least) two situations: 

+ Inside titles `# the elephant [^ in the room]`
+ In figure (and probably other) captions: `Figure: this is a famous [^ or not] picture`.

Actually, the problem is only the `\footnotetext{}` part (which explains why the first code works and the second don't), so that node should be "delayed" right after the caption/figure in TeX, if possible ;)

 .... except if @Karnaj has an idea on the latex side. The error is, for example,

> Argument of \\@caption has an extra }.

TitiAlone:
Hi,

I'm planning on correcting this in the upcoming version of ZMarkdown, could you please provide the expected result from your example above?

TitiAlone:
Okay, just noticed we do have something to prevent this behavior, but... it seems to be misused for `footnoteDefinition`.

The current system, which I can expand to the test case you give, gives the following result:

```latex
\levelOneTitle{hello\textsuperscript{\ref{footnote:1}}\footnotetext[1]{\label{footnote:1} footnote that i like}\protect}
```

Please note the `\protect` at the end which supposedly solves the bug.

Karnaj:
```latex
\levelOneTitle{hello\textsuperscript{\ref{footnote:1}}\protect\footnotetext[1]{\label{footnote:1} footnote that i like}}
```
(I'm also trying to use `\footnotemark` instead of `\label` and `\ref` (it will permit to have the link to the correct page)).

TitiAlone:
Fixed in upcoming version.

PR #413, opened 2020-09-20
Dependencies update & small fixes

### LINKED ISSUE #405: remark-kbd: allow `[[CTRL]]+[[D]]`
### Where

- remark-kbd

### What 

- add option to change fenced chars to `[[` and `]]` like:

```markdown
[[CTRL]]+[[D]]
```

TitiAlone:
My branch was called `minor-work`, because it was supposed to be a quick dependencies update, but remark broke everything (once more). Most of this PR is about adjusting our tests to match these remark changes, and will be stated below with `(U)`; here's a (comprehensive, but likely incomplete) list of what has been done:

- (U) Autolinking is now enabled by remark
    - Even on emails, hence `mailto:demo@example.org` has a new behavior
- (U) Links with space between link and caption are now invalid: `[caption] (link)` is not ok
    - But image are, for some reasons I don't know
- (U) Line breaks are now allowed inside of footnotes
    - To add another paragraph, simply indent it, and it will be part of the footnote
- (U) (quite a huge break) Inline footnotes syntax **has changed**: `[^inline note]` -> `^[inline note]`
- By the way, issue #400 concerning inline footnotes and LaTeX is fixed
- (U) Indent at the beginning of a list renders a codeblock inside the list
- [remark-kbd] Custom characters are now allowed (fix #405 )
- As @Karnaj suggested [here](https://github.com/zestedesavoir/zmarkdown/issues/400#issuecomment-665775099), all `\label` and `\ref` that linked to footnotes were replaced by `\footnotemark`

I think that's it. The PR seems huge, but there are a lot of package-lock additions/deletion.

Also, because of breaking behaviors, this PR should lead to a major version, that will likely not be published directly, as I plan to make another breaking change soon enough.


vhf (approved):
👍 if ZdS is ok with the (breaking) changes :)

TitiAlone:
Will ask, but some of these are deeply integrated in remark, so I guess we do not have a lot of options here :D

TitiAlone:
ZdS validated the changes, and coverage has been improved by two different means:

- exclude all `dist/` files from coverage count;
- add new tests on `common.js`, which was the least tested file.

PR #415, opened 2020-10-01
Allow inverted ranges in hl_lines

### LINKED ISSUE #414: zmarkdown: inverting line number of a range causes a fatal error
The bug only appears with current zmarkdown configuration, you just have to start the server and send provided markdown

# The minimal code

```
```php hl_lines="3-01"
<?php

namespace App\Validator;
```

# observed behaviour

error is raised in `zmarkdown/utils/code-handler.js` at line 70
```
RangeError

Invalid array length
```

# desired behaviour

two possible behaviour : 

- if left limit is greater than right limit, do nothing, 
- swap the range limits if they are inverted

AmauryCarrade:
The parser won't crash if the range is entered inverted by the user, i.e. `hl_lines=2-1` instead of `hl_lines=1-2`. In this example, lines 1 and 2 will be highlighted.

Fixes #414.

#### Notes

- This fixes the problem in the HTML export but not in the LaTeX one where the range stays inverted. Don't know if it's a problem.
- I'm still very new to zmarkdown, so I don't know if I should add a test somewhere for that. I guess so?

### DIFF
--- a/packages/rebber/src/types/code.js
+++ b/packages/rebber/src/types/code.js
@@ -37,6 +37,28 @@ const defaultMacro = (content, lang, attrs) => {
           value = value.slice(1, -1).trim()
         }
 
+        // For hl_lines, we parse the parameters to ensure we don't have
+        // inverted ranges.
+        if (param === 'hl_lines=') {
+          const hlItems = value.split(/, /)
+
+          for (const i in hlItems) {
+            const item = hlItems[i]
+            if (!item.includes('-')) {
+              continue
+            }
+
+            const n = item.split('-').map(n => parseInt(n)).filter(n => !isNaN(n))
+            if (n.length !== 2) {
+              continue
+            }
+
+            hlItems[i] = `${Math.min(n[0], n[1])}-${Math.max(n[0], n[1])}`
+          }
+
+          value = hlItems.join(',')
+        }
+
         localCodeBlockParams[i] = value
       }
     }
--- a/packages/rebber/src/types/code.js
+++ b/packages/rebber/src/types/code.js
@@ -42,8 +42,7 @@ const defaultMacro = (content, lang, attrs) => {
         if (param === 'hl_lines=') {
           const hlItems = value.split(/,| /)
 
-          for (const i in hlItems) {
-            const item = hlItems[i]
+          for (const [i, item] of hlItems.entries()) {
             if (!item.includes('-')) {
               continue
             }
--- a/packages/rebber/src/types/code.js
+++ b/packages/rebber/src/types/code.js
@@ -40,7 +40,7 @@ const defaultMacro = (content, lang, attrs) => {
         // For hl_lines, we parse the parameters to ensure we don't have
         // inverted ranges.
         if (param === 'hl_lines=') {
-          const hlItems = value.split(/, /)
+          const hlItems = value.split(/,| /)
 
           for (const i in hlItems) {
             const item = hlItems[i]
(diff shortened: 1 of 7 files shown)


TitiAlone (commented):
I checked your code, and it seems ok, for HTML at least; for LaTeX, the template requires the arguments *in order* too, so I guess we'll need to reorder. Technically, we could change the order in MDAST directly, so that it applies everywhere, but I am not a huge fan of this solution, I think you could simply handle a specific case in the [`rebber/code`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/rebber/src/types/code.js#L30) handler (rebber is the engine that powers MDAST to LaTeX conversion) ; once your change is done, don't forget to transpile the package using `(npx) lerna run build --scope=rebber`.

About tests, you need to add a test for HTML inside the [`zmarkdown/html-suite.test.js`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/zmarkdown/__tests__/html-suite.test.js#L243) file; we already have a "*code highlight special cases*" block, so I guess you can put it here. For the syntax, have a look at the tests already written, there should not be much to change; to me, you could test for no error (make sure the test fails without your fix), and also for the HTML to contain `data-count="1" class="hll"` or something like that.

If you want to do the LaTeX part too, I do not think we test custom code blocks a lot on `rebber` itself, actually, so maybe adding it to the [`zmarkdown/latex-suite.test.js`](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/zmarkdown/__tests__/latex-suite.test.js#L175) would be enough. It works really like it's HTML counterpart, you could even maybe take the same test, but check for *absence* of `1-3` and *presence* of `3-1`. Please note that the actual behavior with LaTeX is that it **soes not render** the highlighted code, but none of rebber or the template actually fail, so checking for errors on this test would be helpless.

By the way, I forgot to mention you can run tests using the `(npx) jest` command.

AmauryCarrade:
Thanks for these detailed instructions! I added LaTeX support and tests according to what you said. I added a dedicated test in the LaTeX part to avoid cluttering the other one too much.

For the LaTeX part, I made the decision to keep invalid ranges as-is; as example, the following:

``````md
```python hl_lines=2-1,4-
def true():
    return False

True = true()
```
``````

yields the following:

```latex
\begin{CodeBlock}[][1-2,4-]{python}
def true():
    return False

True = true()
\end{CodeBlock}
```

But I can easily _remove_ such ranges if you think it's better (yielding only `\begin{CodeBlock}[][1-2]{python}`, in this example).

packages/rebber/src/types/code.js
```diff
@@ -37,6 +37,28 @@ const defaultMacro = (content, lang, attrs) => {
           value = value.slice(1, -1).trim()
         }
 
+        // For hl_lines, we parse the parameters to ensure we don't have
+        // inverted ranges.
+        if (param === 'hl_lines=') {
+          const hlItems = value.split(/,| /)
+
+          for (const i in hlItems) {
+            const item = hlItems[i]
```
TitiAlone, line 46:
```diff
- for (const i in hlItems) {
-   const item = hlItems[i]
+ for (const [i, item] of hlItems.entries()) {
```

TitiAlone (approved):
Code approved; if you could simply do the small improvement I proposed, and undo spaces changes in files.

AmauryCarrade:
Here we go.

TitiAlone:
Woah, didn't think Babel would add that level of craziness into the machine. Anyway it's merged and ready for the next version of ZMarkdown (10.0.0). Thanks!

AmauryCarrade:
I wasn't expecting this either. :D But it's harmless. Thanks!

Issue #384, opened 2020-02-24
[rebber] Missing dependencies in package.json


thomasdondorf:
Just tried to give rebber a try. Looks like it is missing two dependencies in the `package.json` file:

- `clone`
- `trim-lines`

After manually installing them I was able to get it to run. Used version was `5.0.0`.

TitiAlone:
Hello there,

Thanks for reporting the issue, it will be solved in the next release of `rebber` coming soon. Feel free to ask if you have any questions about `rebber` or it's plugins.

TitiAlone:
Released today: `rebber@5.1.0`

thomasdondorf:
Nice, thanks for the fast feedback. Just to let you know: In the end, I didn't end up using `rebber`, mainly because of the missing TypeScript support. Also, I needed something to work directly on the AST.

Looks like a cool project, though :)

artragis:
Do you know how we could support typescript?

TitiAlone:
We need to provide typings I suppose.

thomasdondorf:
@artragis You could either put the type definition files into your project (like [Unified did it](https://github.com/unifiedjs/unified/blob/master/types/index.d.ts)) or provide them via [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped). Or you switch from JavaScript to TypeScript, but I guess that's not an option ;)

LukeCarrier:
@artragis, @TitiAlone would you accept a PR adding TypeScript declarations for `remark-abbr`?

TitiAlone:
Please do; adding them to only some packages in the repo doesn't really make sense IMO, but I'm interested in seeing how typing declarations work. Do you think adding them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) would make sense?

LukeCarrier:
Awesome, I'll raise a PR for just one package for now and we can go from there. Thanks!

PR #419, opened 2020-10-24
remark-abbr: TypeScript type definitions

LukeCarrier:
This PR is a follow on from the query in https://github.com/zestedesavoir/zmarkdown/issues/384#issuecomment-706429418

I've added TypeScript definitions (ambient modules) for the `remark-abbr` module, and ensured that they're published as part of `npm pack`.

I wasn't quite sure how best to test the definitions as part of your dev/test workflow, but I've included a configuration for `dtslint` which runs as part of `test`. Feedback welcome!


TitiAlone:
I hadn't look at it in depth yet, but I was expecting these to go in the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repository instead of this repository. I mentioned it in [my reply on your comment](https://github.com/zestedesavoir/zmarkdown/issues/384#issuecomment-707665027), but maybe didn't made it clear. Do you think there's a need to have them inside this repo instead of using DefinitelyTyped?

LukeCarrier:
It was easier to get them working here, as to get them into DefinitelyTyped we. We'd to get a couple of dependencies (`remark-abbr` and `unified`) whitelisted in their linting tool, as by default only packages in DT itself are available.

I'm happy to move them into DT if you prefer; sorry for glossing over that point.

TitiAlone:
If that's not a problem for you, I'd prefer them in the DefinitelyTyped reporistory. Don't know what would work better: should I review the typings here, so that you can put them directly on DT, or you can mention me in your PR on their side to get a review.

Anyway, I'll keep this PR opened to keep track of it, and close it when typings are merged into DT.

LukeCarrier:
I'll link the PRs in DT once they're open 🙂

LukeCarrier:
I've raised the two PRs:

* DefinitelyTyped/DefinitelyTyped#49069 contains the new type definitions.
* microsoft/DefinitelyTyped-tools#132 adds `remark` and `unified` to the `package.json` whitelist, and should fix the failing CI job. This is necessary because both packages ship their type definitions in their NPM packages rather than via DT.

TitiAlone:
I see the two PRs have been merged. Is everything fine or do we need to do something here?

LukeCarrier:
No action necessary 😃

TitiAlone:
Thanks once more for the work done!

PR #420, opened 2020-12-19
Add quizz to zmarkdown and rebber stringification

artragis:
- new custom block
- on latex the quizz is splitted into two :
  - the quizz itself as a list
  - the correction as an hidden block with checked list boxes
- implement checked listboxes for latex, uses amssymb package

Nota : on my brand new environment, I cannot run test even if I followed README.md steps.


packages/zmarkdown/config/latex/prepare-quizz.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = processQuizzFactory
+const clone = require('clone')
+function processQuizzFactory (ctx) {
+  const correctionTitle = ctx.correctionTitle || 'Correction'
+  return function quizzCustomBlockVisitor (node, index, parent) {
+    node.type = 'neutralCustomBlock'
+    const correction = clone(node)
+    correction.type = 'sCustomBlock'
+    correction.children[0].children[0].value = correctionTitle
```
StaloneLab, line 11: Please check that `correction.children[0].children[0]` exists and do something if it doesn't, we had a lot of problems with direct indexing like this in the past. Maybe simply revert the change on `node.type` and return?

packages/zmarkdown/config/latex/prepare-quizz.js
```diff
@@ -0,0 +1,22 @@
...
+    parent.children.splice(index + 1, 0, correction)
+    const bodyChildren = node.children[1].children
```
StaloneLab, line 13: Same here: what happens if `node.children[1]` doesn't exist? Please add at least a small check.

StaloneLab (changes requested):
Look globally fine to me; I made some comments on the code, but they're mostly small improvements.
The biggest problem is that your file is at the wrong place: LaTeX preprocessors are not expected to be in the configuration directory.

BTW, if you still experience problems running the tests, I'd be happy to hear about it, feel free to open an issue.

🎅

packages/zmarkdown/config/latex/index.js
```diff
@@ -11,7 +11,9 @@ const rebberConfig = {
       'sCustomBlock',
       'secretCustomBlock',
     ]),
-    heading: require('rebber-plugins/dist/preprocessors/headingVisitor'),
+
+    heading:          require('rebber-plugins/dist/preprocessors/headingVisitor'),
+    quizzCustomBlock: require('./prepare-quizz'),
```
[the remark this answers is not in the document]
artragis: My point is : logically rebber-pluggin is a package with all the stringifiers to use to translate our custom nodes into latex.

Here, we do have a specific code for zmd, not for the translation of remark-custom-block.

packages/zmarkdown/config/latex/prepare-quizz.js
artragis: logically we ensure that quizz has a title so the children[0].children[0] MUST be there.

packages/zmarkdown/config/latex/prepare-quizz.js
artragis: This one, I will make the test but once more, we configured custom block parsing so that it forces to have a title and a block

packages/zmarkdown/config/latex/index.js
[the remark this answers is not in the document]
StaloneLab: `rebber-plugin` is a bit of a mess anyway, we already have these kind of preprocessors in it.
Look at `spoilerFlatten` for instance, it's ZMd-specific but in `rebber-plugins` anyway.

To me, `rebber-plugin` gathers everything needed on ZMd-side for `rebber`; altough it may be debatable, I don't think it's such a bad thing.

packages/zmarkdown/config/latex/prepare-quizz.js
StaloneLab: Well, I may be a bit too paranoid here; my point was: it cost nothing to add a test and we already had problems with theses kind of elements that "we thought would be here but are not" :D
Maybe we could skip the tests.

packages/zmarkdown/config/latex/index.js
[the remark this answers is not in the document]
artragis: Ok, I will move it this afternoon

StaloneLab (approved):

StaloneLab:
Thanks for having taken time to implement this, it should be released with ZMd v10 soon enough.

artragis:
\o/

PR #421, opened 2020-12-24
Default to `text` for code highlighting

StaloneLab:
The title is pretty self-explanatory: currently, when highlighting a code block without specifying a language, `lowlight` tries to guess a best match using regexes, but this procedure was recently found slow and sometimes unstable, so I suggest to remove it.


artragis:
For me this is good.

StaloneLab:
Just noticed it is what we already do for LaTeX, so there shouldn't be too much problem applying this fix ;).

StaloneLab:
Closed because was integrated into #422

PR #423, opened 2021-01-03
[remark-disable-tokenizers] Fix #412

### LINKED ISSUE #412: remark-disable-tokenizers breaks the block tokenizers
This is reported in the remark tracker: https://github.com/remarkjs/remark/issues/529. remark-disable-tokenizers is broken in the same way.

StaloneLab:
fixes: #412 

`remark-disabled-tokenizers` replaces a tokenizer by `() => false` instead of `() => true`, and also adds a dummy locator for inline tokenizers.


artragis:
Looks good to me

artragis (approved):

PR #426, opened 2021-02-01
Add a manifest renderer

StaloneLab:
Add an endpoint to the ZMarkdown server that takes a manifest (ie. file describing the full architecture of a content, see below) and renders it as HTML, LaTeX or EPUB.

## What's a manifest, anyway?

[This gist](https://gist.github.com/StaloneLab/bd3ed923043f29b385c0481a0e436ec8) shows an example of a manifest as provided by `zds-site` (`manifest-raw.json`) and the corresponding manifest required by ZMarkdown (`manifest-arduino.json`). Basically, it describes a content and it's internal structure.

Currently, `zds-site` sends every extract to ZMarkdown, and assembles it itself, in an unparallelized manner, leading to a slow rendering process. The goal is to speed it up, and also to assemble VFiles correctly (that is, error messages, statistics, etc) to get a coherent result.


packages/zmarkdown/server/controllers/manifest.js
```diff
@@ -0,0 +1,116 @@
...
+
+          template(opts, (e, doc) => {
+            if (e) return sendResponse(e)
+
+            // Replace the content, and discard metadata, which have no meaning in LaTeX
+            Object.assign(assembledVfile, {contents: doc, data: {}})
+          })
+        }
+
+        sendResponse(null, assembledVfile)
```
vhf, line 112: You have to wait for `template`'s callback to be done before sending the response, otherwise it might send `assembledVfile` before `Object.assign` (L108) happens.

packages/zmarkdown/utils/vfile-assembler.js
```diff
@@ -0,0 +1,53 @@
+const listConcatUnique = (a, b) => {
+  if (typeof a !== 'undefined' || typeof b !== 'undefined') {
```
vhf, line 2: You probably meant `&&`. There's also a simpler way of implementing this if it does what I think it does, something like (not tested):

```
const listConcatUnique = (a, b) => {
if (Array.isArray(a) && Array.isArray(b)) {
  return Array.from(new Set(a.concat(b)))
}
return []
}
```

StaloneLab: Thanks, fixed in the new version, and your code worked without a change :+1:

packages/zmarkdown/server/controllers/manifest.js
StaloneLab: Refactored using `Promise.all` and fixed this.

vhf (approved):

PR #427, opened 2021-02-17
[zmarkdown] Fix titles depth on manifest rendering

StaloneLab:
Small fix on titles depth not handled correctly.

Opening a PR because I want to see if tests pass. They fail on my side for no obvious reason...


StaloneLab:
Ok, that's what I was thinking: I have a problem on my side.

PR #430, opened 2021-03-06
rebber-plugins: fix quizz latexification

artragis:
Send the answer list in the non-hidden block.


StaloneLab (approved):

StaloneLab:
Will be included in the `10.1.0` version, thanks for the PR!

PR #437, opened 2021-05-29
[remark-images-download] Do not throw error when HTTP status code is 301

Situphen:


StaloneLab:
Thanks for this PR, unfortunately I do not think it can be merged as is. When a `301` or `302` code is returned to the Node `http` agent, it doesn't follow the redirection. Hence, with your changes, no error is thrown but the image inserted in the resulting content is invalid (with an empty file downloaded I think). In fact, it seems the default http module doesn't support following redirects by default, so if you want to fix it, you'll likely need to:

- implement a way to follow redirects, for instance by using the [follow-redirect](https://www.npmjs.com/package/follow-redirects) package from NPM;
- add a test to make sure everything works as expected (301 are redirected and the image. In fact, this can be quite tricky for this module: firstly, you need to add a way to generate 301 in the [mock server](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-images-download/__mock__/server.js). Secondly, add a test in the [corresponding file](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-images-download/__tests__/index.js);
- I'm also thinking of a potential infinite redirects problem (looping on 301), that will need careful attention. If you feel inspired, a test might also be added for it;
- finally, but that should be a formality, run `npm run build -- --scope=remark-images-download` to run Babel on your changes.

Hope I'm not forgetting anything, you'll likely encounter problems I haven't even thought of anyway. Feel free if you have any questions, you're not restricted to empty messages :upside_down_face: .

artragis:
Warning, any "follow-redirect" implementation MUST define a max-hop attribute to avoid redirection bombing. For us it should be set to 1 as we mainly want to follow http -> https.

StaloneLab:
Yes, that's exactly, I think what you are referring to was what I meant by "potential infinite redirects problem" ;)

Situphen:
I first thought this would be an easy fix but it clearly isn't. I will see if I can work to fix this or if it's easier to fix the content themselves.

StaloneLab:
Closing for inactivity. Do not hesitate to reopen either this PR or a new issue for the problem.

PR #442, opened 2021-07-22
[remark-ping] Improve parsing to support Unicode

StaloneLab:
Rewrite almost entirely the parser in `remark-ping` to get better support for Unicode characters in non-fenced pings. For instance, to ping `Moté`, the most natural way would be `@**Moté**`, which didn't work before this PR.

The detection of a ping has been changed as follow:

- before, we used a RegEx, that matched only `\w` characters;
- now, the package contains a complete, RegEx-independant parser, that gets the end of a ping matching characters from Unicode categories *White Space*, *Close Puncutation*, *Final Punctuation*, *Initial Punctuation*, *Open Punctuation* and *Other Punctuation*.

These changes should also make it easier to switch the package to `micromark` soon, as it already consist of a state-machine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


StaloneLab:
Of course CI cries because of Node 10... Time to remove it, *farewell*!

PR #478, opened 2022-11-15
rebber-plugins/math: fix some escaping issues

### LINKED ISSUE #477: [rebber-plugins]Escapement issue with command inside command
When using rebber with rebber-plugin, especially the math plugin, we have issues with escaping

# Input

```
$$\Delta \varphi = 2 \pi f \Delta t_{orange/bleu} = -\frac{\pi}{4}$$
```

# Actual Output

```
$$\Delta \varphi = 2 \pi f \Delta t_{orange/bleu} = -\frac{ {4}}$$
````

# Expected output

```
$$\Delta \varphi = 2 \pi f \Delta t_{orange/bleu} = -\frac{\pi}{4}$$
```

Note : the issue is also present in the full display version of math. 

The `\frac{\pi}{4}` is transformed into `\frac{{4}}`

### PR DESCRIPTION
**Pull request by artragis**
fix #477

### CODE REVIEW

**Review by StaloneLab — COMMENTED** [review]
Hi, thanks for the proposed fix, and the added unit test. Unfortunately, I don't think your solution will work for every possible command: for instance, consider `\%`, which is in the list of allowed commands. This command does match your RegEx (the `%` character will pass `[^\w_-]`): a command can contain non-alphanumeric characters. More generally, finding a solution in the form of a negative search does not seem that easy to me because the command `\{` is in the allowed list. If you have an idea on how to fix, then I would be happy to review again, otherwise the test will be merged anyway.

**Comment by artragis** [comment]
Let me see that next week. 

I have an idea but no time to implement it now.

**Comment by StaloneLab** [comment]
Hi, have you had some time to work on the problem? I am planning to work on it next week if you have no time, so keep me informed 😉 .

**Comment by artragis** [comment]
I think I can do it by the end of the week. If I can't I will keep you informed on thursday

**Comment by artragis** [comment]
Now, I have a better algorithm that passes the tests.

I also added the `\%` command.

**Comment by artragis** [comment]
sorry for all the mess, I did not expect the global test suite to have more tests that were not expecting the right result.

File: packages/rebber-plugins/dist/preprocessors/mathEscape.js [hunk 1]
```diff
@@ -2,6 +2,23 @@
...
+  } // handle the case of \\
+
+
+  var currentChar = nodeLine.charAt(index);
+
+  if (currentChar === '\\' && currentIntendedCommand === '') {
+    return false;
+  }
+
+  return endOfCommandChars.indexOf(currentChar) !== -1;
```
StaloneLab, line 19: Use `includes` instead of `indexOf` if not using the index.

StaloneLab (changes requested):
Some examples do not work as expected I think. For instance consider the two following cases (one example per line):

```latex
a=\mathrm{b}\\\outputrce{\inputrce{/bin/bash} hello hello}
\\cdleft{\frac{1}{2}}
```

While reviewing your work, I got the idea to eat *at most* two slashes at the beginning of the command, so I am issuing a commit on your PR (if I can) to show a solution I have in mind, but as you have worked on it, I would be happy to have your opinion on the changes. If you think everything is fine I will arrange merging.

artragis:
If you have a PR you can send it, I will certainly review it and merge it if it works.

artragis:
Everything seems to be good on my side.

StaloneLab:
I just merged the changes by squashing your commits. Thanks for the help!

PR #484, opened 2022-12-19
Fix/482 843

artragis:
Fix and add test for those two cases


packages/rebber-plugins/src/preprocessors/katexConstants.json
```diff
@@ -1 +1 @@
...
...,"\\Beta","\\bull","\\Chi","\\clubs","\\cnums","\\Complex","\\Dagger","\\diamonds","\\empty","\\Epsilon","\\Eta","\\exist","\\harr","\\hArr","\\Harr","\\hearts","\\image","\\infin","\\Iota","\\isin","\\Kappa","\\larr","\\lArr","\\Larr","\\lrarr","\\lrArr","\\Lrarr","\\Mu","\\natnums","\\Nu","\\Omicron","\\plusmn","\\rarr","\\rArr","\\Rarr","\\real","\\reals","\\Reals","\\Rho","\\sdot","\\sect","\\spades","\\sub","\\sube","\\supe","\\Tau","\\thetasym","\\weierp","\\Zeta","\\argmin","\\argmax","\\plim","\\bra","\\ket","\\braket","\\Bra","\\Ket","\\angln","\\blue","\\orange","\\pink","\\red","\\green","\\gray","\\purple","\\blueA","\\blueB","\\blueC","\\blueD","\\blueE","\\tealA","\\tealB","\\tealC","\\tealD","\\tealE","\\greenA","\\greenB","\\greenC","\\greenD","\\greenE","\\goldA","\\goldB","\\goldC","\\goldD","\\goldE","\\redA","\\redB","\\redC","\\redD","\\redE","\\maroonA","\\maroonB","\\maroonC","\\maroonD","\\maroonE","\\purpleA","\\purpleB","\\purpleC","\\purpleD","\\purpleE","\\mintA","\\mintB","\\mintC","\\grayA","\\grayB","\\grayC","\\grayD","\\grayE","\\grayF","\\grayG","\\grayH","\\grayI","\\kaBlue","\\kaGreen","\\ce", "\\\\", "\\nabla", "\\{", "\\}", "\\,", "\\[", "\\}"]
```
StaloneLab, line 1: All the commands you added already exist in the file (check before with a quick search), some of them even appear twice. I will regenerate the constants file before releasing to remove the duplicates, but can you remove yours already?

The only exception (symbol not already in the file) is the `\[` which doesn't seem to be parsed by KaTeX (from what I can see in the live demo), so no changes should be required on this file.

packages/rebber-plugins/src/preprocessors/mathEscape.js
```diff
@@ -1,6 +1,6 @@
 const katexConstants = require('../../src/preprocessors/katexConstants.json')
 
-const endOfCommandChars = ['\\', '[', ']', '{', '}', ' ', '(', ')', '\t', '\n']
+const endOfCommandChars = ['\\', '[', ']', '{', '}', ' ', '(', ')', '\t', '\n', '^', '_', ',']
```
StaloneLab, line 3: This array seems to just grow bigger and bigger :smile: 

If we have one more problem, I will find another way to detect end of command.

StaloneLab (changes requested):
Fine by me, only two requested changes:

- could you remove your additions in `katexConstants.json`, it should still work (see comment);
- could you compile your changes, using `npm run build -- --scope=rebber-plugins`.

artragis:
Seems to be ok by now.

StaloneLab (approved):
Fine by me, I think it will be released as a bugfix version, see the other PR.

PR #501, opened 2023-12-09
[zmarkdown] Fix getting processus status in Munin controller

philippemilink:
Since status `online` is coded as `0`, `0` was evaluated as `false` and thus the fallback value `3` was taken. Now, we use a ternary operator to test if the status is in the dictionnary to map to a number.

The concerned URL is http://localhost:27272/munin/status.


StaloneLab:
Perfect, thanks!

PR #502, opened 2023-12-16
[zmarkdown] Set graph_vlabel for processus status in Munin controller

philippemilink:
Otherwise Munin complains:
```
[WARNING] 1 lines had errors while 8 lines were correct in data from 'config zmd_status'
```
The part of the label `0=online, 1=stopped, 2=errored` could be generated automatically from the `status` object, but that's would maybe be over-engineered (and I don't know how to do it properly!).

I also added some property to the graph.

Here is how it looks:
zmd_status-day

The concerned URL is http://127.0.0.1:27272/munin/config/status.

### DIFF
--- a/packages/zmarkdown/server/controllers/munin.js
+++ b/packages/zmarkdown/server/controllers/munin.js
@@ -7,7 +7,8 @@ module.exports = isConfig => (req, res) => {
     status: {
       graph: [
         'graph_title Process Status',
-        'graph_vlabel',
+        'graph_vlabel Status (0=online, 1=stopped, 2=errored)',
+        'graph_args --lower-limit 0 --upper-limit 3',
       ],
       fields: (name) => [
         `${name}.label ${name}`,


StaloneLab:
I don't like that much how the graph is renderer... Do you know if there is another way to show this legend? Munin does not seem to support custom things for axis labels. Maybe we should simply skip a line after "Status", or not display it at all?
Anyways, I agree what you suggest would be better than nothing, as I always thought this graph was not working anymore since it showed 0 all the time.

Also, to generate the legend automatically from the `status` object, you could use:

```javascript
Object.entries(status).map(([label, state]) => `${state}=${label}`).join(', ')
```

Of course, that would require moving the `status` object at the beginning of the block, before definition of the graph.

LikaKavkasidze:
Please see commit [zmarkdown@425a5395d1de9b6ac5743c5a4d9a4ae8af9f3372](https://github.com/zestedesavoir/zmarkdown/commit/425a5395d1de9b6ac5743c5a4d9a4ae8af9f3372), based on your work. Closing the PR, but feel free to comment if the proposed solution does not suit your needs.

PR #504, opened 2023-12-29
fix: fix webpack 5 config for live demo

### LINKED ISSUE #494: zmarkdown: fix live demo
Following the upgrade to Node 18/20, Webpack was upgraded to version 5 in order to be able to build the live demo. Even though some changes were made to keep the compilation working, it seems that the bundled files are wrong, because nothing seems to work.

I, as a maintainer, and knowing almost nothing about Webpack, am not ready to spend any more time on this issue. As a consequence, the live demo will stay in it’s current (non-working) state until one of these two conditions is met:

1. all our imports are made statically, which should happen when deploying `next`;
2. someone is ready to spend some time fixing it.

Vayel:
Fix #494 

# Test

1. `npm run build-demo`
2. Open `packages/zmarkdown/public/index.html`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,6 @@ Currently, all the plugins provided only work for remark versions **lesser than*
 
 This project uses [Jest][jest] for testing. It is recommended to use the locally installed version using `npx`, and run Jest in watch mode when developing `npx jest --watch --notify` (`--notify` sends desktop notifications when tests run).
 
-### Development
-
-Read [`packages/zmarkdown/README.md`](zmarkdown-dev)
-
 ### Useful commands
 
 * `npm run test` : tests all packages.
--- a/README.md
+++ b/README.md
@@ -166,4 +166,3 @@ This project uses [Jest][jest] for testing. It is recommended to use the locally
 [typographic-colon]: https://github.com/zestedesavoir/zmarkdown/tree/master/packages/typographic-colon#typographic-colon--
 [typographic-permille]: https://github.com/zestedesavoir/zmarkdown/tree/master/packages/typographic-permille#typographic-permille--
 [zmarkdown]: https://github.com/zestedesavoir/zmarkdown/tree/master/packages/zmarkdown#zmarkdown--
-[zmarkdown-dev]: https://github.com/zestedesavoir/zmarkdown/tree/master/packages/zmarkdown#dev-build
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ Currently, all the plugins provided only work for remark versions **lesser than*
 
 This project uses [Jest][jest] for testing. It is recommended to use the locally installed version using `npx`, and run Jest in watch mode when developing `npx jest --watch --notify` (`--notify` sends desktop notifications when tests run).
 
+### Development
+
+1. `npm run build-demo`
+1. Open `packages/zmarkdown/public/index.html` in your browser
+
 ### Useful commands
 
 * `npm run test` : tests all packages.
--- a/README.md
+++ b/README.md
@@ -27,8 +27,7 @@ This project uses [Jest][jest] for testing. It is recommended to use the locally
 
 ### Development
 
-1. `npm run build-demo`
-1. Open `packages/zmarkdown/public/index.html` in your browser
+Read [`packages/zmarkdown/README.md`](zmarkdown-dev)
 
 ### Useful commands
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -25,6 +25,11 @@ Currently, all the plugins provided only work for remark versions **lesser than*
 
 This project uses [Jest][jest] for testing. It is recommended to use the locally installed version using `npx`, and run Jest in watch mode when developing `npx jest --watch --notify` (`--notify` sends desktop notifications when tests run).
 
+### Development
+
+1. `npm run build-demo`
+1. Open `packages/zmarkdown/public/index.html` in your browser
```
StaloneLab, line 31: The command `build-demo` was mostly made for internal use (ie. to release the demo on Github), I am not in favor of documenting it inside of the main README, but if you want you can update the `packages/zmarkdown` README, which contains the procedure to create development and production builds (at the end of the file).

packages/zmarkdown/webpack.config.js
```diff
@@ -1,3 +1,4 @@
+const webpack = require('webpack');
```
StaloneLab, line 1: We use `eslint` to ensure consistency of the code style, and require no semicolon at the end of lines, please remove it, that should make the tests work.

StaloneLab (changes requested):
Thanks a lot for your work, I cannot understand anything to Webpack, but I tested it locally and the live demo is working again!

I left two minor comments to be addressed, and I'll merge your changes right after.

README.md
Vayel: How do you test your code locally if not by running the demo? As a new contributor, I was stuck on that.

Starting the zmd server and making a POST request appears a bit laborious to me.

StaloneLab: ZdS, and all other users of ZMarkdown I am aware of, are using the server, therefore I rely exclusively on the server, and mostly use tests for checking problems. Implementing a new functionality is done using TDD most of the time; we have more than 1000 tests all over the project, so one is likely to fail is something broke. Occasionally, I use some tools [I have developed](https://github.com/StaloneLab/zmarkdown-experiments/tree/master/publisher) to compile a full tutorial, allowing to do some basic integration testing, but unit tests are usually enough.

To clarify what the client build is for: the live demo only, not development purposes, nor use in a production environment. We might have this kind of use someday, but it is not the current use (indeed, since the client build require some Mo to be loaded). Why do we keep it then? Because it has almost zero maintenance cost (did not break in more than five years, it is the first time), and some people use it occasionally to check Markdown syntax or see if a bug is indeed due to ZMd.

New developers are encouraged to use the server, for two quite simple reason: our tests will not run without a working server, and the server is what is used anyway. Actually, tests start the server when launched. For this PR, you had no choice but to deploy the demo (of course), however if you intend to work on something else, here is the recommended workflow:

1. write at least one test for your feature/bugfix;
2. write your code and build it with `npm run build`;
3. launch tests with `npm run test`;
4. repeat 2/3 until all tests (including yours) pass.

So regarding the idea of adding these directives to the main README, I still stand opposed, but I could accept something like "More documentation is located in the package/zmd README, especially parameters used for the server and usage of a client build" (not exactly this, since the sentence is weird).

StaloneLab (commented):
There are no more linting problems, so apart from the `README` part, to which I answered, this is ready to be merged.

Unit testing is still failing because of `package-lock.json`, but no need to worry about this, I'll fix it once merged, as it can prove quite tricky not to mess up since we use a monorepo.

README.md
Vayel: Thanks for the answer! It'd be nice to have a short tutorial in the README on how to add a simple example feature step by step.

StaloneLab: I guess you messed something up in your latest commits because Github only shows an addition of a link to the end of the main README which is not used inside of the content.

Vayel: Fixed

StaloneLab:
Ok, I thought you wanted to add something to the README but let's merge as is. I am ready to merge other changes you suggested if you want to work on them, such as:

- adding a link to the `package/zmd` README in the main one;
- adding some developer documentation in the README.

You might want to do this last one while working on something else, if you want to involve in the project.

StaloneLab:
Live demo is indeed working: https://zestedesavoir.github.io/zmarkdown/. Thanks again for your work!

PR #505, opened 2024-01-05
[zmarkdown] Lazy-load images

### LINKED ISSUE #496: HTML export: should we add a lazy-loading attribute to all images and iframes?
The HTML Living Standard now includes a `loading` attribute for [images](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading) and [iframes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#loading), allowing them to be loaded only when they are close to the viewing window. This is an easy way to reduce the initial rendering time of some huge web pages full of images and iframes - think of content packed with JSFiddles on Zeste de Savoir.

The `loading="lazy"` attribute is [now supported everywhere](https://caniuse.com/loading-lazy-attr), except for iframes on Firefox (support is coming with Firefox 121, current nightly).

Adding a `loading="lazy"` attribute to all images and iframes might therefore be a good idea. I’d be open to contributing, if this sounds like a good idea to anyone other than myself.

LikaKavkasidze:
Fixes #496 

Lazy-loading for iframes was added by commit https://github.com/zestedesavoir/zmarkdown/commit/28b2196031cefd1fb8da480bf031fff2814c7187 , this PR aims at lazy-loading all images for ZMarkdown.

Since parsing of images is done by remark directly, implementation was done using a postprocessor, which adds an attribute `loading=lazy` for all images, except emoticons. Since these are guaranteed to be lightweight and served from our servers, I considered they could be eagerly loaded (see condition in postprocessor).

As it is the case with all postprocessors, this one can be easily disabled with a configuration option (`lazyLoadImages`). Since this feature should be ignored by browser not supporting it, I feel like we do not need a major version for enforcing this setting in the default configuration.

### DIFF
--- a/packages/zmarkdown/__tests__/__snapshots__/html-suite.test.js.snap
+++ b/packages/zmarkdown/__tests__/__snapshots__/html-suite.test.js.snap
@@ -84,13 +84,13 @@ exports[`code highlight special cases supports linenostart 1`] = `
 `;
 
 exports[`images become figures: does not apply when a caption is present 1`] = `
-"<figure><img src=\\"http://example.com\\" alt=\\"foo\\"><figcaption>foo</figcaption></figure>
-<figure><img src=\\"http://example.com\\"><figcaption>Caption</figcaption></figure>
-<figure><img src=\\"http://example.com\\" alt=\\"foo\\"><figcaption>Caption</figcaption></figure>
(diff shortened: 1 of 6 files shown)


artragis (approved):

LikaKavkasidze:
Thanks @artragis , I am releasing a new version `zmarkdown@11.4.0` with the recent changes.

PR #506, opened 2024-01-07
Properly support code-blocks in remark-grid-tables

damageboy:
Cell contents is not preprocessed to selectively trim
whitespace in such a way that leading whitespace inside of a markdown
code-block is not trimmed beyond the indentation "set" by the opening
of said code-block. Previously all leading whitespace was trimmed,
making code-block indentetaion within a table-cell impossible.

### DIFF
--- a/packages/remark-grid-tables/src/index.js
+++ b/packages/remark-grid-tables/src/index.js
@@ -376,6 +376,49 @@ function extractTableContent (lines, linesInfos, hasHeader) {
   return table
 }
 
+function processCellLines (lines) {
+  const trimmedLines = []
+  let inCodeBlock = false
+  let leadingWhitespace = 0
+  let leadingWhitespaceRegex = null
+
+  for (let i = 0; i < lines.length; i++) {
+    let line = lines[i]
+
+    // Trim the end of the line
+    line = line.trimEnd()
+
+    // Check if we're entering or exiting a code block
+    if (line.trim().startsWith('```')) {
+      inCodeBlock = !inCodeBlock
+
+      // If we're entering a code block, remember the amount of leading whitespace
+      if (inCodeBlock) {
+        // Set how much whitespace to remoev from lines
+        // inside the code-block
+        leadingWhitespace = line.match(/^ */)[0].length
+        leadingWhitespaceRegex = new RegExp(`^ {0,${leadingWhitespace}}`)
+      }
+
+      // Remove leading whitespace from the opening/closing code-block
+      // statement as well
+      line = line.replace(leadingWhitespaceRegex, '')
+    } else if (inCodeBlock) {
+      // If we're *already* in a code block, trim the start of the line
+      // by the amount of leading whitespace
+      line = line.replace(leadingWhitespaceRegex, '')
+    } else {
+      // We're not in a code block, trim the start of the line
+      line = line.trimStart()
+    }
+
+    // Replace the line in the array
+    trimmedLines.push(line)
+  }
+
+  return trimmedLines
+}
+
 function generateTable (tableContent, now, tokenizer) {
   // Generate the gridTable node to insert in the AST
   const tableElt = {
@@ -408,8 +451,10 @@ function generateTable (tableContent, now, tokenizer) {
       }
       for (let c = 0; c < row._cells.length; c++) {
         const cell = row._cells[c]
+        const trimmedLines = processCellLines(cell._lines)
+
         const tokenizedContent = tokenizer.tokenizeBlock(
-          cell._lines.map((e) => e.trim()).join('\n'),
+          trimmedLines.join('\n'),
           now
         )
         const cellElt = {
--- a/packages/remark-grid-tables/dist/index.js
+++ b/packages/remark-grid-tables/dist/index.js
@@ -463,6 +463,46 @@ function extractTableContent(lines, linesInfos, hasHeader) {
   return table;
 }
 
+function processCellLines(lines) {
+  var trimmedLines = [];
+  var inCodeBlock = false;
+  var leadingWhitespace = 0;
+  var leadingWhitespaceRegex = null;
+
+  for (var i = 0; i < lines.length; i++) {
+    var line = lines[i]; // Trim the end of the line
+
+    line = line.trimEnd(); // Check if we're entering or exiting a code block
+
+    if (line.trim().startsWith('```')) {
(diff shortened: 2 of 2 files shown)


LikaKavkasidze:
I’ll have a proper look later, but at first sight it seems like a very good first PR, thanks!
Two minors things I can say for sure: I ran the linter on your code (see Github Actions), and it does not seem to adhere to our rules, especially regarding semicolons, could you please have a look and try to fix it?
Also, we still have (although we started dropping it) some transpilation of code for older JS engines, could you run the transpiler using `npm run build -- --scope=remark-grid-tables`.

If you would like to learn a bit more about the project, to make further contributions for instance, I would also not be against a test for the behavior you’re trying to fix.
I am ready to help on the subject should you encounter any problem, but I do not require this test to merge the PR, it will be merged as is if you do not want to work on the project any further.

Finally, it is kind of rare for us to receive PRs from people outside of the project, so may I ask how you encountered this issue? Are you using the ZMarkdown server or our plugins for some kind of project?

damageboy:
> Two minors things I can say for sure: I ran the linter on your code (see Github Actions), and it does not seem to adhere to our rules, especially regarding semicolons, could you please have a look and try to fix it?

I'll fix linting and get GH Actions green, thanks!

I'm trying to use the remark-grid-tables (which are basically restructured text tables AFAICT?) in a docusaurus MDX documentation site since I need the colspan/rowspan feature set as well as actual paragraphs in tables... (code-blocks, lists).

I thought I'd give this remark "plugin" a run, and it performed rather well, which is also how I discovered that it "eats" away important whitespace in code block... hence my attempt at this PR

packages/remark-grid-tables/src/index.js
```diff
@@ -376,6 +376,47 @@ function extractTableContent (lines, linesInfos, hasHeader) {
...
+        leadingWhitespace = line.match(/^\s*/)[0].length
+      }
+
+      // Remove leading whitespace from the opening/closing code-block
+      // statement as well
+      line = line.substring(leadingWhitespace)
+    } else if (inCodeBlock) {
+      // If we're *already* in a code block, trim the start of the line
+      // by the amount of leading whitespace
+      line = line.substring(leadingWhitespace)
```
LikaKavkasidze, line 407: Let's say I have the following Markdown:

```markdown
+------+----------------------+
| cell | `code                |
|      | with line break`     |
+------+----------------------+
|  ```text                    |
| line 1                      |
| line2                       |
|```                          |
+------+----------------------+
```

The output will have the first character of the code stripped for each line. I am quite unsure of what could be done in that case, maybe counting the number of whitespaces every time and if it is lesser than the number currently saved, update it?

packages/remark-grid-tables/src/index.js
```diff
@@ -376,6 +376,47 @@ function extractTableContent (lines, linesInfos, hasHeader) {
...
+    } else {
+      // We're not in a code block, trim the start of the line
+      line = line.replace(/^\s+/, '')
```
LikaKavkasidze, line 410: Please use the `trimStart` method for Strings: `line = line.trimStart()`

LikaKavkasidze (changes requested):
Fine overall. I left a few comments, one of them is quite a big issue which I do not know how it can be addressed easily...

packages/remark-grid-tables/src/index.js
damageboy: Yeah, you're right. But I don't see this as hard to address...
It "just" requires to use a regexp that removes "that many" space chars from the untrimmed string, if there are less... It should probably do nothing (let the user face their own doing).

If there are more, then it's all good..

damageboy:
Let me know if you get a chance to re-review. In my testing this is ready.
If you want me to add many small test cases for these weird edge-cases, just let me know.

LikaKavkasidze:
I will for sure, I saw your changes yesterday but couldn't review right away. It will be done this week ;)

packages/remark-grid-tables/src/index.js
LikaKavkasidze: What's hard is to find which behavior is to be implemented, not to choose one ;)

That's also why I needed some time to review your changes: I checked how it is done in [`micromark-extension-gridtables`](https://github.com/adobe/micromark-extension-gridtables), which will likely be the plugin we will use when switching to the new micromark system. The way they implemented it (which actually delegates to `micromark` core), gives the same results as your approach, so let's do it like this!

packages/rebber-plugins/__tests__/fixtures/mix-7.fixture.md
```diff
@@ -2,8 +2,8 @@
 | cell | `code                |
 |      | with line break`     |
 +------+----------------------+
-|  ```text                    |
-| line 1                      |
-| line2                       |
+|```text                      |
+|line 1                       |
+|line2                        |
```
LikaKavkasidze, line 7: I would like to keep this test as is, so updating the fixture is not the way to go. You should however update the snapshot, which requires reverting changes on this file and running `npm test -- packages/rabber-plugins -u` at the root of the repository.

LikaKavkasidze (changes requested):
Fine overall, please update snapshot(s) instead of changing the test content, and I will merge.

packages/rebber-plugins/__tests__/fixtures/mix-7.fixture.md
LikaKavkasidze: Since I see you asked if you should add tests: you can, but I'm not requiring it. I know which tests I'd like to have (ie. micromark compatibility), so I'll add a few if you don't after merging.

damageboy: Thanks for the comments, coming back to this now, will do as you asked

damageboy:
Hey, I've done as requested and it looks like I shouldn't have updated the test in the first place.

All test pass locally without any modification now on my local machine (all 1051 snapshots)

LikaKavkasidze (approved):
Fine by me, thanks for the PR, a new version will be released before the end of the month, can be earlier if you urgently need it.

damageboy:
Thanks for all th help and patience with a JS n00b :)

LikaKavkasidze:
One day late, but I just published a new version including your changes: [`remark-grid-tables@2.2.1`](https://github.com/zestedesavoir/zmarkdown/releases/tag/remark-grid-tables%402.2.1).

PR #507, opened 2024-03-28
[remark-images-download] add a user-agent to requests

philippemilink:
Hello,

This PR solves a problem observed on zestedesavoir.com: downloading images from upload.wikimedia.org failed with a HTTP 403 error, but not on beta.zestedesavoir.com...

After reading [this article](https://meta.wikimedia.org/wiki/User-Agent_policy) and live-patching this package on zestedesavoir.com, I can confirm the lack of user-agent was responsible for these HTTP 403 errors.

It will be hard to test, just believe me that this patch is now used in production. And speaking of test, I don't think we can add a test to cover this case...

I used [this article](https://zestedesavoir.com/billets/4640/mon-installation-domotique/) to debug it (the HomeAssistant logo is an image from Wikimedia).

### DIFF
--- a/packages/remark-images-download/README.md
+++ b/packages/remark-images-download/README.md
@@ -107,7 +107,7 @@ All options are optional.
 
   If not provided, local images will not end up in `downloadDestination`.
 
-## example
+## Example
 
 ```markdown
 Two small images:
--- a/packages/remark-images-download/__mock__/server.js
+++ b/packages/remark-images-download/__mock__/server.js
@@ -12,6 +12,16 @@ fd.on('data', (data) => {
   }
 })
 
+app.use((req, res, next) => {
+  /* Some websites (like Wikimedia) may refuse requests without User-Agent
+   * header, so mimic their behaviour: */
+  if (!req.get('User-Agent')) {
+    res.writeHead(403)
+    res.end()
+  }
+  next()
+})
+
 const bloat = Buffer.alloc(10 * 1024)
 
 app.get('/stream-bomb', (req, res) => {
--- a/packages/remark-images-download/dist/index.js
+++ b/packages/remark-images-download/dist/index.js
@@ -272,11 +272,19 @@ function plugin() {
   // Rejects with an error if headers are invalid.
   var initDownload = function initDownload(url) {
     return new Promise(function (resolve, reject) {
+      var packageInfo = require('../package.json');
+
       var parsedUrl = new URL(url);
       var proto = parsedUrl.protocol === 'https:' ? https : http;
-      var req = proto.get(parsedUrl, {
-        timeout: httpRequestTimeout
-      }, function (res) {
+      var reqOptions = {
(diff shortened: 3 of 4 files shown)


LikaKavkasidze:
Hi :wave: ,

before merging, you need to build the package using `npm run build -- --scope=remark-images-download`, as mentioned in the `README.md`. Before doing that, since you linked the Wikimedia page, could you follow their convention on how to choose the User-Agent:

> The generic format is `<client name>/<version> (<contact information>) <library/framework name>/<version> [<library name>/<version> ...]`. Parts that are not applicable can be omitted.
> 
> If you run an automated agent, please consider following the Internet-wide convention of including the string "bot" in the User-Agent string, in any combination of lowercase or uppercase letters. This is recognized by Wikimedia's systems, and used to classify traffic and provide more accurate statistics.

So I guess something like `remark-images-download bot/3.0.3 (https://github.com/zestedesavoir/zmarkdown)` would do it (adding bot and version). The version might be parsed directly from the `package.json` file.

Regarding tests, I am a bit surprised we do not tests for headers received by the mock server used for testing. I think it would be nice to have at least something checking for headers, if not on the mock server, maybe on the request sent. This will avoid unexpected header changes when updating our dependencies, or forgetting some of them for the micromark migration.

philippemilink:
> before merging, you need to build the package using `npm run build -- --scope=remark-images-download`, as mentioned in the `README.md`. 

Well, it is not clear we have to run this command after every code modification in packages. And it feels a bit odd to me to generate a dist file and commit it, IMO this should be done only when preparing the release. Anyway, this PR is not about this, I did it.

> So I guess something like `remark-images-download bot/3.0.3 (https://github.com/zestedesavoir/zmarkdown)` would do it (adding bot and version). The version might be parsed directly from the `package.json` file.

Done.


> Regarding tests, I am a bit surprised we do not tests for headers received by the mock server used for testing. I think it would be nice to have at least something checking for headers, if not on the mock server, maybe on the request sent. This will avoid unexpected header changes when updating our dependencies, or forgetting some of them for the micromark migration.

I changed the mock server in a way it rejects requests without User-Agent header.

LikaKavkasidze:
> it feels a bit odd to me to generate a dist file and commit it, IMO this should be done only when preparing the release.

Maybe it is, but it would be easy to forget some packages when preparing the release, and we cannot simply build everything on every release because sometimes it creates a lot of garbage on other packages.

Anyway, thanks for the PR, I merged it, will release a new version with the changes next week most likely.

LikaKavkasidze:
Just released a new version with your changes, see [`remark-images-download@3.0.4`](https://github.com/zestedesavoir/zmarkdown/releases/tag/remark-images-download%403.0.4) and [`zmarkdown@11.4.2`](https://github.com/zestedesavoir/zmarkdown/releases/tag/zmarkdown%4011.4.2).

PR #512, opened 2024-04-26
[zmarkdown] Allow using custom element for iframes

StaloneLab:
Add an option to ZMarkdown server, `hide_iframes`, which will replace all `iframe` elements by `hidden-frame` elements, for all iframe providers configured in the `remark-iframes` configuration file.


StaloneLab:
@Situphen : should I merge this? What has been decided regarding RGPD-compliance?

Situphen:
Hi Stalone, nothing has been decided yet with the team. I'll push this subject and get back to you with an answer ;)

philippemilink:
We finally discussed this topic [during our technical meeting](https://zestedesavoir.com/forums/sujet/11108/reunion-des-dev-de-zeste-de-savoir/?page=9#changer-l%C3%A9l%C3%A9ment-html-iframe-lors-du-rendu-zmd-dune-int%C3%A9gration-vid%C3%A9o), and the conclusion is we would be happy with this new feature!

I just added it could be enabled/disabled through an option (so we can update zmd in zds-site even if we are not ready in zds-site to handle the consent request), but I see it is already done!

So go ahead and merge! (although I didn't test it and we are not ready in zds-site to use this feature).

Issue #514, opened 2024-09-17
Future of remark-abbr


richardTowers:
Hi folks,

As discussed in https://github.com/zestedesavoir/zmarkdown/issues/416 there are a number of remark-plugins which are not yet compatible with remark@13, and users are encouraged to port their own plugins.

I've put together [a remark-abbr plugin](https://github.com/richardTowers/remark-abbr/). The implementation isn't entirely compatible with zmarkdown's version (e.g. it doesn't support [expandFirst](https://github.com/zestedesavoir/zmarkdown/blob/6508e7e4b0a286d23c0b2ae242a12a6b9aaf6389/packages/remark-abbr/README.md#optionsexpandfirst)), and it hasn't been used much yet so there could be some bugs. For the most part though, I'm fairly happy with it.

Right now, it's published to NPM under my namespace [@richardtowers/remark-abbr](https://www.npmjs.com/package/@richardtowers/remark-abbr).

I'm interested in your views on options for merging the two implementations. I think the options are:

1. Keep them separate, with my implementation remaining namespaced (do nothing)
2. Move my implementation into the zmarkdown monorepo, release a new major version of remark-abbr, and have zmarkdown take on maintenance of the plugin
3. Transfer ownership of the NPM package, release my version as a new major version of remark-abbr, and have me take on maintenance of the plugin

I'd be happy with any of these options. It would be nice for users looking for remark-abbr to get a version that supports the latest remark, but my personal use case is already met so we don't _have_ to merge them. If (3), I'm probably not going to do a particularly stellar job of providing support for the plugin, but then hopefully the maintenance requirements should be small. On the other hand, if (2) I feel bad dumping a bunch of code on you folks to maintain.

Opinions welcome!

StaloneLab:
Hello, maintainer of the repository here.

When I commented on this topic of the migration in the past, my opinion was mainly that there is no reason for the migration to take a long time, so I was mostly against releasing the already-migrated plugins, especially considering that this would be more complicated for us to maintain two codebases (we still depend on the old ones ourselves, until every single package in this monorepo is migrated).

At a point, noting that some people might be interested in contributing, I then said that if they did, they I would be ready to review their merge request, and release a new version of the package. This has not changed, and will likely not: anyone wanting to contribute here is welcome, and I will even take the time I don't have to help.

[...]

PR #515, opened 2024-09-21
Add remark-abbr to tests and re-record snapshots

richardTowers:
The only difference in the snapshots seems to be the presence of `\\` backslashes before quote characters.

I'm a little confused by the output, not being all that familiar with jest, but I suspect this is some minor difference in the way that rehype-stringify works and that we don't need to worry about it?

This is just laying a small bit of groundwork, before I attempt to get remark-abbr working with the new version of remark

See also #514

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
   },
   "scripts": {
     "pretest": "lerna run pretest --scope zmarkdown",
-    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules DEST=/tmp jest packages/remark-kbd packages/remark-iframes packages/remark-ping packages/micromark-extension-kbd packages/micromark-extension-iframes packages/micromark-extension-ping packages/micromark-extension-sub-super",
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules DEST=/tmp jest packages/remark-abbr packages/remark-kbd packages/remark-iframes packages/remark-ping packages/micromark-extension-kbd packages/micromark-extension-iframes packages/micromark-extension-ping packages/micromark-extension-sub-super",
     "lint": "eslint .",
     "posttest": "lerna run posttest --scope zmarkdown",
     "build": "lerna run build",
--- a/packages/remark-abbr/__tests__/__snapshots__/index.js.snap
+++ b/packages/remark-abbr/__tests__/__snapshots__/index.js.snap
@@ -27,22 +27,22 @@ exports[`compiles to markdown 3`] = `
 *[HTML]: HyperText Markup Language"
 `;
 
-exports[`empty object does not break with references in their own paragraphs 1`] = `"<p>Here is a test featuring <abbr title=\\"A B C\\">abc</abbr> and <abbr title=\\"D E F\\">def</abbr></p>"`;
+exports[`empty object does not break with references in their own paragraphs 1`] = `"<p>Here is a test featuring <abbr title="A B C">abc</abbr> and <abbr title="D E F">def</abbr></p>"`;
 
 exports[`empty object no reference 1`] = `"<p>No reference!</p>"`;
 
 exports[`empty object passes the first regression test 1`] = `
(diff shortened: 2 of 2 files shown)


StaloneLab:
I'm way too surprised that the tests work (apart from these backslashes, I'll check). It seems that actually the plugin might work out of the box on new remark versions: it relies only on `unist-util-visit` which operates only on AST and does not require anything special. I think you would have noticed if it was the case, what am I missing? Or maybe you wanted to use the low-level micromark API?

StaloneLab:
Hum I see unrelated linting problems, I fixed on `next`, you should be able to rebase (I forced push so maybe cherry-pick is easier). On top of that, there are some other changes to be done which we do while transitioning to `next`. I think the easiest way is to merge and I'll do the changes right after. They are stupid changes anyway, purely workflow, not really relevant to what you are trying to achieve.

richardTowers:
Haha 🤣 

I didn't try the version on the `next` branch, I just tried the version published to npm, discovered that it didn't work, and then found the issue talking about support for remark@13 and assumed it would need significant rework.

I didn't even bother to read the code properly, I just jumped straight in to building my own plugin from scratch. So it's a suprise to me that it works too! (but only a suprise because I made bad assumptions). 

As you say, if the plugin does everything at the AST level, it shouldn't be bothered by the change in parser. So hopefully a minimal bit of tidy up and we're all good.

My odyssea of dodgy assumptions so far:
- I assumed that this plugin needed a lot of rework, so I didn't spend a lot of time understanding it, I just jumped in to rewriting it
- I initially assumed that The Right Way To Do It would be to parse everything at the micromark level, and then do pretty minimal work on the AST. I had a [discussion about that with the micromark folks](https://github.com/orgs/micromark/discussions/181), which concluded that micromark is missing a couple of extension points that would enable parsing a new kind of definition (like abbreviation definitions are), and that I'd be best doing some of the work at the AST level.
- I assumed that I had to parse abbreviation definitions at the micromark level (rather than finding them in an AST transform), so I wrote [a micromark extesnion for that](https://github.com/richardTowers/remark-abbr/tree/main/dev/lib/micromark-extension-abbr), and just handled the abbreviations themselve by transforming the AST

I'll port some of my test suite over, and we can see if there's any reason to bother with micromark, or if this covers all the edge cases well enough already.

StaloneLab:
To be clear: I am not at all advocating for our plugin, I think the way it currently works by basically using Regexes on the AST is quite weird. So if your extension has a more "micromark-compatible" way of doing things, I think it would be nice to implement anyway.

Also, the version published on NPM should be quite (if not exactly) similar to what we have in the `next` branch today. So if you say this didn't worked, then I am definitely missing something. I think I have a test setup somewhere for remark, so I'll test in the upcoming days (maybe today?).

I read your code quite rapidly, so I might have misunderstood things, I will read properly on MR, but from what I understand:

- micromark is used to parse abbr definitions (i.e. `*[W3C]: thing`), while;
- a transformer is used to replace occurences of the abbr inside of text nodes.

Is this right? If not, can you correct me?

Regarding this MR, it looks ready to merge. I will once the current state of `remark-abbr` on remark > 13 will have been clarified (i.e. I will have tested the extension to see what doesn't work).

richardTowers:
Yeah, that's exactly correct. Micromark parses the abbreviation definitions, and then a transformer splits text nodes into `[text, abbr, text, abbr, text]` etc.

StaloneLab:
So… I tested, and in fact we do that today: using an (old-style) `remark` tokenizer and then a transformer that replaces inside of text nodes. Maybe you could get inspired by the old code. In any case I'll look at it when reviewing yours.

Merging this PR now, and will get the repo fully ready for your changes.

StaloneLab:
And since I forgot to answer about the backslash changes: I do not think we care much for now, if it becomes reversed after your change you might change it again, I do not think I encountered it before when migrating plugins so I am not sure if expected or not.

StaloneLab:
Ok so now tests do not run: they were actually still run with the old version of remark. You can start integration whenever you want, everything is ready on my side. I also disabled or fixed other failing tests so your objective is to get back to CI green check.

PR #516, opened 2024-11-09
Update remark-abbr to work with remark@next

richardTowers:
This is a rough and ready port of richardTowers/remark-abbr over the zmarkdown.

I've tried to stick as closely to the original tests as possible - most of them pass with only some minor changes to the fixtures. I've brought over my own tests (converted to Jest) for the micromark extension, but I haven't brought in my remark-abbr tests just yet.

I've added support for expandFirst to get as much of the test suite passing as possible.

I've dropped about half the type annotations I had in the original - I'm not sure if fully typing each plugin is a goal of zmarkdown, so it didn't seem worth it to fix all the comments. If it's not a goal, I'll remove the remaining type annotations.

The remark-abbr plugin is also making a half hearted attempt to keep track of positional information, but I'm fairly confident it's wrong so I should probably strip it out. I think the micromark syntax extension gets the positions mostly right, but transforming the tree and splitting out the text nodes makes the positions very confusing (particularly with expandFirst).


StaloneLab:
Hello, very sorry I have little time right now, but I saw your PR and will have a look!

richardTowers:
Hello! Absolutely no rush at all, it's not blocking anything for me. Take your time!  (But also thanks for having a look)

packages/remark-abbr/lib/index.js
```diff
@@ -1,152 +1,247 @@
...
+      return {
+        abbr,
+        start,
+        end,
+        prevChar: textNode.value[start - 1],
+        nextChar: textNode.value[end + 1]
+      }
+    })
+    .filter((match) =>
+      // We don't want to match "HTML" inside strings like "HHHHTMLLLLLL", so check that the
```
StaloneLab, line 31: Looks like this actually existed before, since currently HTML works, but I have no idea how it works. Will have to check, because abbr appears in tree.

packages/remark-abbr/lib/index.js
```diff
@@ -1,152 +1,247 @@
...
+      // surrounding characters are either undefined (i.e. start of string / end of string)
+      // or non-word characters
+      [match.prevChar, match.nextChar].every(
+        (c) => c === undefined || /^\W$/.test(c)
```
StaloneLab, line 35: I disagree with `\W` because it does not work outside of ASCII characters. For example try the plugin with abbr `*[НПО]: неправительственная организация` (russian for NGO) and input text `НННПООО` (same thing as what you did for `HTML`, I doubled the letters). The trailing `Н` (not latin H) and leading `О` (not latin O) are taken for non-word characters. But they are word characters, in another language. I suggest to use a proper plugin instead, I think we use something similar somewhere, search for `unicode` in project dependencies.

StaloneLab (commented):
Erh, I had wrote a comment but apparently these are not saved by Github. Overall okay, I left a few comments here and there, but have another point that disturbs me: currently, the micromark plugin does not work as-is. I usually try to do it, to allow people without AST (including me sometimes), ie. using `micromark` outside of `remark`, to get things done. I am however not sure if this is currently possible with micromark. Have you tried it and got no success at it?

I will do a second review later, after your answers and fixups, especially for `remark-abbr`, since I will need to run the code to understand a few points. Finally, to answer your questions:

- we do not *care* about type annotations *per-se*, and use them nowhere in the repository, but if we can have them for free for this plugin, and it requires almost no work on our side, then I see no reason to drop them I guess;
- about positional information, if you noticed them wrong, then please remove them, it is not worth the fight except if you really want it. If somebody wants it done someday, then he will make a PR. I do not want to bother checking them if you already know they are not right currently (but will if you fix everything, even though I say it is not needed). I would say these are perfect candidates for test-driven development (ie. write first a test describing the positions you expect for a few cases) if you want to work on it.

Thanks again for your work, this is looking like we may be able to ship quite easily.

StaloneLab:
PS: do not bother about tests on Node 18 I would say, especially since they fail only there.

